PR #8, opened 2014-05-26
Node sizes

jimfulton:
This is an alternative to the feature.configure_tree_bucket_sizes branch for controlling node sizes.

You control maximum node sizes by subclassing a BTree class and overriding max_internal_size and/or max_leaf_size.


jimfulton:
WRT increasing default sizes:

+1 :) How much?

No evolution of existing data is needed.

Krilivye:
Suggestion for defining how much:

Look at size of lot of existing production BTree => mean it.
:)
:+1:

jimfulton:
Did you just volunteer to come back with a recommendation? :)

Krilivye:
Rule of thumb in our team looks like "Correct" Large BTree should be no more than 100K.
Else start rework on data design. This is in the current default configuration. (250Max/buckets of 30).
I will benchmark different size next week :)

PR #11, opened 2015-04-07
Fix setting _p_changed for small pure-Python BTrees.

jamadden:
When working on [zopefoundation/zc.zodbdgc#1](https://github.com/zopefoundation/zc.zodbdgc/pull/1#discussion-diff-27853357), we found a bug in the pure-Python BTree implementation. An `elif` that should have been its own `if` prevented the second and subsequent additions to a BTree from setting the `_p_changed` flag, right up until the first bucket needed to split. 

This can cause data-loss for small BTrees.

This PR adds a test case and should fix the issue (the comment is copied from the C implementation).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ coverage.xml
 *.egg
 dist
 .eggs/
+.dir-locals.el
--- a/BTrees/_base.py
+++ b/BTrees/_base.py
@@ -883,11 +883,16 @@ class _Tree(_Base):
                 max_size = self.max_leaf_size
             if child.size > max_size:
                 self._grow(child, index)
-        elif (grew is not None and
-              child.__class__ is self._bucket_type and
-              len(data) == 1 and
-              child._p_oid is None
-              ):
+
+        # If a BTree contains only a single bucket, BTree.__getstate__()
+        # includes the bucket's entire state, and the bucket doesn't get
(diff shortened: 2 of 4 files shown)


tseaver:
This fix LGTM.  I plan to experiment with the new test (verifying failure in `py27-pure`, success in `py27`) before merging.  Too bad the darned test suite takes so long to run.

jamadden:
It definitely fails for me without the change and works with it, but of course feel free to confirm :) 

```
[master]$ nosetests BTrees.tests.test_OLBTree:OLBTreeTest.testAddTwoSetsChanged
F
======================================================================
FAIL: testAddTwoSetsChanged (BTrees.tests.test_OLBTree.OLBTreeTest)
  File "BTrees/BTrees/tests/common.py", line 1121, in testAddTwoSetsChanged
    self.assertTrue(t._p_changed)
AssertionError: False is not true

[master]$ cd fixed
[fixed]$ nosetests BTrees.tests.test_OLBTree:OLBTreeTest.testAddTwoSetsChanged
.
```

On my machine at least, the Python 2.6 tests run almost twice as fast as many of the other versions (90s vs 193s for Py2.7). Travis runs them all about the same...I wonder why that is...

tseaver:
I forgot that the `py27` tests would exercise the pure-Python stuff, too.  I can definitely verify the fix.

tseaver:
Thank you!

PR #12, opened 2015-04-07
Make all the tests in the w_zodb Tox environment run.

jamadden:
Prior to this PR, a large number of tests in the `w_zodb` environment would fail. Most of the errors came from `BTrees.tests.common:Base._closeRoot` trying to close a connection joined to a transaction. The remainder came from `BTrees.tests.testConflict` trying to use a non-MVCC-aware FileStorage from two different connections, leading to a "Duplicate tpc_begin calls for same transaction" error. This PR makes all tests in the `w_zodb` environment run (unfortunately that's not tested by Travis).

### DIFF
--- a/BTrees/tests/testConflict.py
+++ b/BTrees/tests/testConflict.py
@@ -504,37 +504,35 @@ class NastyConfictFunctionalTests(ConflictTestBase, unittest.TestCase):
 
     @_skip_wo_ZODB
     def testConflictOfInsertAndDeleteOfFirstBucketItem(self):
-        """ testConflictOfInsertAndDeleteOfFirstBucketItem
-        Recently, BTrees became careful about removing internal keys
-        (keys in internal aka BTree nodes) when they were deleted from
-        buckets. This poses a problem for conflict resolution.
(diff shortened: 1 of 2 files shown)


BTrees/tests/testConflict.py
```diff
@@ -478,7 +504,7 @@ def testConflictWithOneEmptyBucket(self):
 
     @_skip_wo_ZODB
     def testConflictOfInsertAndDeleteOfFirstBucketItem(self):
-        """
+        """ testConflictOfInsertAndDeleteOfFirstBucketItem
```
tseaver: This part doesn't look right, somehow

Actually, our coding standard would be to replace the docstring with a comment (docstrings in testcases mess up the error reporting).

tseaver:
Other than the docstring -> comment note, LGTM.

BTrees/tests/testConflict.py
jamadden: Hehe, the docstring did mess up the error reporting so I echoed the test name to the top of it and forgot about it. I can turn it into a comment.

tseaver:
Thanks!

PR #13, opened 2015-05-06
Fix setting _p_changed when removing from small Python BTrees/TreeSets.

jamadden:
Sadly, the fix for #11 was incomplete (sorry) and failed to handle removal, which bit us testing our application under PyPy. This PR handles removal and adds test cases for it.

Example test failures before the fix:
```
======================================================================
FAIL: testRemoveInSmallMapSetsChanged (BTrees.tests.test_OOBTree.OOBTreePyTest)
  File "//BTrees/BTrees/tests/common.py", line 1179, in testRemoveInSmallMapSetsChanged
    self.assertTrue(t._p_changed)
AssertionError: False is not true

======================================================================
FAIL: testRemoveInSmallSetSetsChanged (BTrees.tests.test_OOBTree.OOTreeSetPyTest)
  File "//BTrees/BTrees/tests/common.py", line 1421, in testRemoveInSmallSetSetsChanged
    self.assertTrue(t._p_changed)
AssertionError: False is not true

```

Twice, out of many test runs, I saw the following test failure for Python2.7. I'm pretty sure it's unrelated:
```
======================================================================
ERROR: test_LP294788 (BTrees.tests.testBTrees.BugFixes)
  File "//BTrees/BTrees/tests/testBTrees.py", line 355, in test_LP294788
    id = trandom.choice(list(ids.keys()))
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/random.py", line 275, in choice
    return seq[int(self.random() * len(seq))]  # raises IndexError if seq is empty
IndexError: list index out of range

```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@
 4.1.3 (unreleased)
 ------------------
 
-- TBD
+- Fix _p_changed when removing items from small pure-Python
+  BTrees/TreeSets. See:
+  https://github.com/zopefoundation/BTrees/issues/13
 
 
 4.1.2 (2015-04-07)
@@ -13,8 +15,9 @@
 - Suppress testing 64-bit values in OLBTrees on 32 bit machines.
   See:  https://github.com/zopefoundation/BTrees/issues/9
 
-- Fix _p_changed for small pure-Python BTrees.
-  See   https://github.com/zopefoundation/BTrees/issues/11
+- Fix _p_changed when adding items to small pure-Python
+  BTrees/TreeSets. See:
+  https://github.com/zopefoundation/BTrees/issues/11
 
 
 4.1.1 (2014-12-27)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@
 ------------------
 
 - Fix _p_changed when removing items from small pure-Python
-  BTrees/TreeSets. See:
+  BTrees/TreeSets and when adding to Sets. See:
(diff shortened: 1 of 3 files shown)


jamadden:
It turns out that plain `xSet` objects have the same issue as the `Tree` versions did in #11. I added a fix and tests to this PR.

CHANGES.rst
```diff
@@ -4,7 +4,9 @@
 4.1.3 (unreleased)
 ------------------
 
-- TBD
+- Fix _p_changed when removing items from small pure-Python
+  BTrees/TreeSets and when adding to Sets. See:
+  https://github.com/zopefoundation/BTrees/issues/13
```
tseaver: Don't we also re-fix #11?  Should be mentioned here (so that folks who experienced the bug in 4.1.2 know to expect that it is fixed).

tseaver:
I'm good to merge this once the changelog is updated.

CHANGES.rst
jamadden: Issue 11 was about adding to the Tree versions; that didn't change here. This issue handles adding to the non-Tree Sets, as well as removing from Trees. I'm not sure how to reword the change entry to make it more clear, but I'm open to any suggestions.

tseaver: OK, I do see that it mentions the sets.

tseaver:
Released via: https://pypi.python.org/pypi/BTrees/4.1.3

PR #16, opened 2015-09-15
Ensure that we don't overlook errors in first 'PyObject_RichCompareBool' call

### LINKED ISSUE #15: Tests break on Python 3.5
```
ValueError: incomparable                                                                                [0/1943]

During handling of the above exception, another exception occurred:

  File "/tmp/nix-build-python3.5-BTrees-4.1.4.drv-0/BTrees-4.1.4/BTrees/tests/testBTrees.py", line 426, in testF
oo
    t[DoesntLikeBeingCompared()] = None
  File "/tmp/nix-build-python3.5-BTrees-4.1.4.drv-0/BTrees-4.1.4/BTrees/tests/testBTrees.py", line 416, in __cmp
__
    raise ValueError('incomparable')
SystemError: <class 'ValueError'> returned a result with an error set

```

tseaver:
Python 3.5 turns such cases into SystemErrors.

See: https://bugs.python.org/issue23571

Fixes #15.

### DIFF
--- a/BTrees/_compat.h
+++ b/BTrees/_compat.h
@@ -27,9 +27,25 @@
 #define TEXT_FROM_STRING PyUnicode_FromString
 #define TEXT_FORMAT PyUnicode_Format
 
-#define COMPARE(lhs, rhs) \
-    PyObject_RichCompareBool((lhs), (rhs), Py_LT) > 0 ? -1 : \
-    (PyObject_RichCompareBool((lhs), (rhs), Py_EQ) > 0 ? 0 : 1)
+/* Emulate Python2's __cmp__,  wrapping PyObject_RichCompareBool(),
+ * Return -2/-3 for errors, -1 for lhs<rhs, 0 for lhs==rhs, 1 for lhs>rhs.
+ */
+static inline
+int __compare(PyObject *lhs, PyObject *rhs) {
+    int less, equal;
+
+    less = PyObject_RichCompareBool(lhs, rhs, Py_LT);
+    if ( less == -1 ) {
+        return -2;
+    }
+    equal = PyObject_RichCompareBool(lhs, rhs, Py_EQ);
+    if ( equal == -1 ) {
+        return -3;
+    }
+    return less ? -1 : (equal ? 0 : 1);
+}
+
+#define COMPARE(lhs, rhs) __compare((lhs), (rhs))
 
 
 #else
--- a/BTrees/_compat.h
+++ b/BTrees/_compat.h
@@ -38,11 +38,14 @@ int __compare(PyObject *lhs, PyObject *rhs) {
     if ( less == -1 ) {
         return -2;
     }
+    if (less) {
+        return -1;
+    }
     equal = PyObject_RichCompareBool(lhs, rhs, Py_EQ);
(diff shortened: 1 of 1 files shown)


tseaver:
@jimfulton could you check me here?  The proper error-checking code seemed to hairy to work into an "expression" macro, so I put it in a static inline function and had the macro just call it.

jimfulton:
I'll take a look this weekend.

jimfulton:
I don't understand how this fixes anything.   Nothing is looking for -2 or -3 afaict.  

Furthermore, TEST_KEY_SET_OR, which calls COMPARE calls PyErr_Occurred to detect the error.

Does your change somehow magically make the test pass? If so, I'm baffled as to how it could.

(Your change also forces an unnecessary comparison, in many cases.)

I would try to figure out why PyErr_Occurred doesn't seem to be effective.

(Looking at this after many years, I'm not thrilled with how these macros are laid out. If I worked on them, I'd refactor them a bit to make it a little clearer wtf was going on.)

tseaver:
> I don't understand how this fixes anything. Nothing is looking for -2 or -3 afaict.

The `-2`/`-3` return values are irrelevant for any practical use, because the exception will be raised by the caller.

> Furthermore, TEST_KEY_SET_OR, which calls COMPARE calls PyErr_Occurred to detect the error.  Does your change somehow magically make the test pass? If so, I'm baffled as to how it could.

`PyObject_RichCompareBool` returns `-1` to indicate an error occurred, but the old code ignores that possibility and calls it a second time.  The fix is to avoiding making the second `PyObject_RichCompareBool` call with an error set, which (in Python 3.5) causes a `SystemError`.

> (Your change also forces an unnecessary comparison, in many cases.)

ff4c330 drops the `Py_EQ` comparison if the `Py_LT` one returns true.

jimfulton:
Ah, I see. We were already making 2 calls and the second call was failing. Doh.

So the -2 and -3 are decoys. We're really relying on PyErr_Occurred. Given how down and dirty this code is, I'd rather not add a decoy.  

I suggest instead modifying the original macro to:

    #define COMPARE(lhs, rhs) \
      PyObject_RichCompareBool((lhs), (rhs), Py_LT) != 0 ? -1 : \
      (PyObject_RichCompareBool((lhs), (rhs), Py_EQ) > 0 ? 0 : 1)

And add a comment that 

a) the second comparison is skipped if the first comparison returns 1 (less than) or -1 (error) and that
b) It's up to the caller to detect errors with PyError_Occurred.

domenkozar:
I've applied the patch to 4.1.4 release and it builds and tests pass (I haven't done any other testing). Would be nice to also add py3.5 to travis-ci file.

jimfulton:
Which patch? Tres' or my simpler one?

domenkozar:
@jimfulton Tres

jimfulton:
OK, well, I'd prefer mine because:

- simpler
- less overhead
- less decoys

I guess I'll make a new PR.

jimfulton:
I just made a new PR, https://github.com/zopefoundation/BTrees/pull/17, that fixes this more simply.

Issue #19, opened 2016-01-11
Pickling issues between C and Python implementations.


jamadden:
This grows out of the discussion in zopefoundation/persistent#32. The user had been using the C implementation of BTrees and had many saved pickles. Upon updating dependencies, the user discovered he was using the Python implementation. This was slow, so the user switched back to the C implementation.

This leads to two related issues. The first, most serious, issue is that a BTree pickled by the Python implementation [will only ever unpickle as the Python implementation](https://github.com/zopefoundation/persistent/issues/32#issuecomment-170555885), though a BTree pickled by the C implementation will unpickle as whichever implementation is available (but see below). Because of the tree-like structure of BTree objects, [this can lead to AttributeErrors](https://github.com/zopefoundation/persistent/issues/32#issuecomment-170528248) when the parent object is (re)pickled as the Python implementation and still has children that are pickled as the C implementation.

The second issue I discovered when testing the first issue. An empty C BTree unpickled with the Python implementation isn't initialized correctly, [leading to its own AttributeErrors](https://github.com/zopefoundation/persistent/issues/32#issuecomment-170560244).

In environments like my own organizations, that are slowly rolling out PyPy, it will be common to have developers and production/test environments sharing pickles where everyone is on some combination of PyPy and CPython. Maybe even different production (micro)services will be on PyPy and CPython at the same time. So the first AttributeError could be a real problem.

It seems to me like the two implementations should produce pickles that always result in loading the "best" available implementation, i.e., the Python implementation should not specify the `Py` suffix in pickle names. I think that's the only way to avoid the first issue. Though, there is an argument to be made  that someone might explicitly want to pickle the Python implementation, but I'm not sure why. I'm not sure how to fix this without changing the pickling format in a backwards-incompatible way, but I'm not a pickle expert. I can try to look into that a bit.

I suspect the second issue is easier to fix.

If there's consensus that something needs to be done on this I can do the work to submit a PR.

mgedmin:
This sounds like a serious bug that ought to be fixed.

jamadden:
It *may* be as simple as reversing the names of the Python implementations. Instead of:

```python
class OOBTreePy(BTree): 
   ...

try:
    from ._OOBTree import OOBucket
except ImportError as e: #pragma NO COVER w/ C extensions
    OOBTree = OOBTreePy
    ...
else: #pragma NO COVER w/o C extensions
    from ._OOBTree import OOBTree
    ...
```

Do this:
```python
[...]

PR #20, opened 2016-01-12
Make Python objects pickle like C; unpickling empty BTrees in Python works.

jamadden:
Fixes the issues raised in #19 (originally from zopefoundation/persistent#2). See that issue for the discussion that led to this particular approach. I'm opening the PR now because all the tests are passing for me locally for all versions and I've addressed all the corner cases I can think of.

I realize that altering `__class__` might be controversial. I wouldn't be surprised if others will shortcomings or compatibility issues that need to be addressed, or even if this approach is rejected and it's decided to go with changing the pickle format (just sad :) ).

Adding `_fix_pickle` to the bottom of each module saves a great deal of boilerplate in each module (DRY!), at the expense of (perhaps) some "magic". Personally, I rate DRY above avoiding magic when it comes to simple metaprogramming like this, but I can understand if views differ. (Then again, this approach could be furthered in a subsequent PR to remove more repeated boilerplate found at the bottom of each module to the tune of saving around 40 SLOC per module.)


BTrees/_base.py
```diff
@@ -1494,3 +1546,44 @@ def MERGE_WEIGHT_default(self, value, weight):
...
+
+    for name in 'Bucket', 'Set', 'BTree', 'TreeSet', 'TreeIterator':
+        raw_name = mod_prefix + name
+        py_name = raw_name + 'Py'
+        try:
+            py_type = mod_dict[py_name]
+        except KeyError:
+            if name == 'TreeIterator':
+                # Optional
+                break
```
mgedmin: I'd rather use `continue` here, even though there's no difference in behavior as long as `TreeIterator` is last in the list of things to iterate over.

BTrees/tests/common.py
```diff
@@ -208,6 +211,95 @@ def remove(i):
...
+
+            # If we're testing the pure-Python version, but we have the
+            # C extension available, then the loaded type will be the C
+            # extension but the made type will be the Python version.
+            # Otherwise, they match. (Note that if we don't have C extensions
+            # available, the __name__ will be altered to not have Py in it. See _fix_pickle)
+            if 'Py' in type(made_one).__name__:
+                self.assertTrue(type(loaded_one) is not type(made_one))
+            else:
+                self.assertTrue(type(loaded_one) is type(made_one) is self._getTargetClass(), (type(loaded_one), type(made_one), self._getTargetClass(), repr(dumped_str)))
```
mgedmin: Maybe this long line ought to be wrapped?

BTrees/tests/common.py
```diff
@@ -208,6 +211,95 @@ def remove(i):
...
+        # This doesn't hold for things like Bucket and Set, sadly
+        # self.assertEqual(t, t2)
+
+    def test_pickle_subclass(self):
+        # Issue #2: Make sure our class swizzling doesn't break
+        # pickling subclasses
+
+        global PickleSubclass # XXX: Has to be global to pickle, but this prevents running tests in parallel
+        class PickleSubclass(type(self._makeOne())):
+            pass
```
mgedmin: When you say "prevents running tests in parallel", do you mean `bin/test -j4` and such get random test failures?  That's not nice.

I wonder if you could use something like this instead:
```python
base_class = type(self._makeOne())
class_name = 'PickleSubclassOf' + base_class.__name__
PickleSubclass = type(class_name, (base_class,), {})
globals()[class_name] = PickleSubclass  # has to be global to pickle
```

BTrees/tests/common.py
```diff
@@ -784,6 +876,24 @@ def testPop(self):
...
+    def _getTargetClass(self):
+        # Most of the subclasses override _makeOne and not
+        # _getTargetClass, so we can get the type that way.
+        if type(self)._makeOne is not BTreeTests._makeOne:
+            return type(self._makeOne())
+
+        # Derive the class from the test case name, if not
+        # overridden
+        name = self.__class__.__name__
+        type_name = name[:-4]
```
mgedmin: I'd like a short comment mentioning what suffix the `[:-4]` strips.
```python
type_name = name[:-4]  # e.g. OOBTreeTest -> OOBTree
```
Or, even better, an assertion
```python
assert name.endswith('Test')
type_name = name[:-len('Test')]
```

CHANGES.rst
```diff
@@ -1,6 +1,23 @@
...
+  This prevents interoperability problems and database corruption if
+  both implementations are in use. While it is no longer possible to
+  pickle a Python implementation and have it unpickle to the Python
+  implementation if the C implementation is available, existing Python
+  pickles will still unpickle to the Python implementation (until
+  pickled again). See:
+  https://github.com/zopefoundation/BTrees/issues/19
+
+- Unpickling empty BTrees in a pure-Python environment no longer
+  creates invalid objects that faile with ``AttributeError``.
```
mgedmin: Typo: faile -> fail.

Did this bug have a separate GitHub issue number?

mgedmin:
LGTM.

BTrees/_base.py
jamadden: Good point.

BTrees/tests/common.py
jamadden: The problem with wrapping long lines in `assert` statements is that you lose the context in the traceback, making it harder to match up what failed :( . E.g., instead of something like
```
File "foo" 271
   assertTrue(a = b = c, (a, b, c))
Error: False is not true ("A", "B", "C")
```
You wind up with something like:
```
File "foo" 272
                     (a, b, c))
Error: False is not true ("A", "B", "C")
```

BTrees/tests/common.py
jamadden: It's a theoretical problem I didn't run into in practice. But I think that or a random nonce should fix it,  thanks!

CHANGES.rst
jamadden: Nope, that was all part of #19.

BTrees/tests/common.py
jamadden: Good idea. 

Actually, that whole branch isn't used; I thought I was going to refactor some things to use it, but that turned out to be (1) too much work and (2) out of scope for this PR. So I should just remove it.

BTrees/tests/common.py
mgedmin: That's why I said "maybe" rather than insisting ;-)

jamadden:
CC @tseaver and @jimfulton (who was involved in the discovery of the original bug) in case of additional comments.

jamadden:
gentle ping

BTrees/tests/common.py
ml31415: But as you read the source code more often than the traceback, once the test passes, I guess the readability of the source code outweights the readability of the traceback. Anyways, a matter of taste.

BTrees/_base.py
```diff
@@ -1494,3 +1545,44 @@ def MERGE_WEIGHT_default(self, value, weight):
...
+
+        if not have_c_extensions:
+            # Set FooPy to have the __name__ of simply Foo.
+            # We can't do this if the C extension is available,
+            # because then mod_dict[FooPy.__name__] is not FooPy
+            # and pickle refuses to save something like that.
+            # On the other hand (no C extension) this makes our
+            # Python pickle match the C version by default
+            py_type.__name__ = raw_name
+            py_type.__qualname__ = raw_name # Py 3.3+
```
ml31415: If I understand this fixes correctly, it's basically these five classes, that get their `cls.__name__`, `inst.__class__` and `cls.__qualname__` patched. Wouldn't it be easier, to just set these attributes unconditionally for the Py classes? I suppose I'm missing some gotcha here, but that seems to me the clearer and more straight forward approach.

jamadden: I'm sorry if the comments aren't clear. It's not possible to set these attributes unconditionally because if you do, then pickling the python classes fails: `Pickler.save_global` raises an exception when it finds out that `getattr(module, FooPy.__name__)` (where `FooPy.__name__` has been set to `Foo`) is not the same thing as `getattr(module, 'Foo')`.  This occurs when both the C and Python classes are present and you try to pickle the Python class; maybe this is a corner case, but it arises in the unit tests, and would arise in fixing a corrupted database and maybe elsewhere.

ml31415: Ok, makes sense now. Thanks for clarifying!

jamadden:
It's been about a month so unless there are objections I think I'll probably go ahead and merge this soon.

jimfulton:
Thanks (again) Jason.

Jim

On Sat, Feb 13, 2016 at 6:29 PM, Jason Madden <notifications@github.com>
wrote:

> Merged #20 <https://github.com/zopefoundation/BTrees/pull/20>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zopefoundation/BTrees/pull/20#event-549655441>.
>



-- 
Jim Fulton
http://jimfulton.info

BTrees/_base.py
```diff
@@ -40,6 +40,44 @@ def __init__(self, items=None):
...
+        # Detect the presence of the C extensions.
+        # If they're NOT around, we don't need to do any of the
+        # special pickle support to make Python versions look like
+        # C---we just rename the classes. By not defining these methods,
+        # we can (theoretically) avoid a bit of a slowdown.
+        # If the C extensions are around, we do need these methods, but
+        # these classes are unlikely to be used in production anyway.
+        __import__('BTrees._OOBTree')
+    except ImportError:
+        pass
```
tseaver: Not covered.  See:  http://jenkins.pylonsproject.org/job/BTrees/68/console

BTrees/_base.py
```diff
@@ -40,6 +40,44 @@ def __init__(self, items=None):
...
+
+        @property
+        def _BTree_reduce_as(self):
+            # Return the pickle replacement class for this object.
+            # If the C extensions are available, this will be the
+            # C type (setup by _fix_pickle), otherwise it will be the real
+            # type of this object.
+            # This implementation is replaced by _fix_pickle and exists for
+            # testing purposes.
+            return type(self)
```
tseaver: Not covered.  See http://jenkins.pylonsproject.org/job/BTrees/68/console

BTrees/_base.py
```diff
@@ -1494,3 +1545,44 @@ def MERGE_WEIGHT_default(self, value, weight):
...
+    for name in 'Bucket', 'Set', 'BTree', 'TreeSet', 'TreeIterator':
+        raw_name = mod_prefix + name
+        py_name = raw_name + 'Py'
+        try:
+            py_type = mod_dict[py_name]
+        except KeyError:
+            if name == 'TreeIterator':
+                # Optional
+                continue
+            raise
```
tseaver: Not covered.  See:  http://jenkins.pylonsproject.org/job/BTrees/68/console

BTrees/_base.py
tseaver: 1587-1588 not covered. See:  http://jenkins.pylonsproject.org/job/BTrees/68/console

BTrees/_base.py
jamadden: Indeed. Coverage only runs with C extensions installed, IIRC (it's been awhile). Is there a suggestion on how to cover the other case? Other than the obvious "# no cover"

tseaver: There is a mechanism for sharing coverage data between tox environments.  E.g., Pyramid uses both python2 and python3 for coverage:

  https://github.com/Pylons/pyramid/blob/master/tox.ini#L84

PR #25, opened 2016-03-11
Add Byte flavour BTree

alexkiro:
A possibly useful optimization for storing large number of data that only require smaller values. I.e. `[-127, 127]`


jamadden:
I think if you rebase on current master, you'll get some test failures that were addressed in #20. You'll need to make some minor changes to each of the Python modules to fix them (e.g., [add this](https://github.com/zopefoundation/BTrees/blob/master/BTrees/IIBTree.py#L139) to each one.)

alexkiro:
> I think if you rebase on current master, you'll get some test failures that were addressed in #20

No tests have failed, but that must be because no tests have been added for the functionality. I'll try to add some tests and apply the fixes you indicated. Thanks!

jamadden:
Oh, yeah, right, tests are definitely needed. It should just be a matter of adding a few subclasses of existing base cases. The pattern is pretty clear.

tseaver:
Thanks for the patch!  When I first read the description, I thought the `B` stood for *bytes*, not for a single *byte*.  I'm not sure I see the point of using the BTrees machinery for single-byte keys:  surely it would be more efficient (in both space and time) to use a single, 256-element array?

As far as values (the `IBBTree` case), the space savings are miniscule compared to `IIBTree`.

If I've misunderstood somehow, please feel free to clarify / reopen.

alexkiro:
> When I first read the description, I thought the B stood for bytes, not for a single byte. I'm not sure I see the point of using the BTrees machinery for single-byte keys: surely it would be more efficient (in both space and time) to use a single, 256-element array?

I see your point here using a single byte as a key doesn't make much sense. TBH these were only implemented for completion. I'm only actually using the IB implementation here. 

> As far as values (the IBBTree case), the space savings are miniscule compared to IIBTree.

I wouldn't really consider them to be minuscule. We had issues with IIBTree using up huge amounts of memory which was solved by switching to the IBBTree one. For example a quick benchmark using 100 millions values ([file here](https://gist.github.com/alexkiro/6ef0fd23ac0f080043ce)):

```
$ time python memt.py II
18.54296875 MiB (pmem(rss=19443712, vms=91566080, shared=8310784, text=3772416, lib=0, data=10788864, dirty=0))
Using II, adding 100000000 values
1858.109375 MiB (pmem(rss=1948368896, vms=2020151296, shared=8376320, text=3772416, lib=0, data=1939374080, dirty=0))

real	1m55.666s
user	1m54.016s
sys	0m1.124s
```

```
$ time python memt.py IB
18.671875 MiB (pmem(rss=19578880, vms=91561984, shared=8450048, text=3772416, lib=0, data=10784768, dirty=0))
Using IB, adding 100000000 values
1298.44921875 MiB (pmem(rss=1361522688, vms=1433247744, shared=8515584, text=3772416, lib=0, data=1352470528, dirty=0))

real	1m57.764s
user	1m56.420s
sys	0m0.676s
```

We save almost 600 MB. When reaching ~1 Billion values this gain becomes even more significant. @tseaver do you think the IB one is worth including upstream?

tseaver:
Looks reasonable to add the `IB` version, then.

tseaver:
@alexkiro can you diagnose / fix the [failing](https://travis-ci.org/zopefoundation/BTrees/builds/115797606) `IBBTree` tests?

jamadden:
Is anyone (@alexkiro ) interested in moving this PR forward? It has a failing test, and some conflicts to resolve by rebasing on master.

jamadden:
There seems no interest in moving this forward over the last three years. There are so many conflicts, I think it best to close. Feel free to reopen if there’s interest!

PR #28, opened 2016-03-25
Feature.configure tree bucket sizes

Krilivye:
Hi tseaver, I'm very interested in finalizing this feature.
I had a hard time dealing with massive write (+500 000 up to 2M) and have observed at least a non linear time on insert due to Tree growth.
After dealing localy with your solution on OOBtree (only) for testing I am very motivated to make this option available in standard Btree package :+1:


tseaver:
@Krilivye I asked for feedback on that branch just over two years ago on the ZODB mailing list:

  http://comments.gmane.org/gmane.comp.web.zope.zodb/12706

and heard nothing.  At this point, I would need at least an OK from @jimfulton to merge the branch before working on resurrecting it.

jimfulton:
Sorry, I should have closed this.

I Implemented this another way:

  https://github.com/zopefoundation/BTrees/pull/8

Basically, you configure Bucket sizes using custom BTree classes.

Tres sent me a private email with a proposal and a branch link to which I responded with extensive comments. I never got a response to my comments.

jimfulton:
FTR, here are some responses to this when it as a proposal and branch:

On Tue, Mar 11, 2014 at 5:46 PM, Tres Seaver <tseaver@palladion.com> wrote:
> -----BEGIN PGP SIGNED MESSAGE-----
> Hash: SHA1
>
> Background
> ==========
>
> In order to tune performance on large sites, I'd like to be able to
> experiment with different bucket / tree sizes.  For instance, we have
> sites where the number of loads for buckets is much too high for some
> kinds of requests:  increasing the bucket size could reduce the
> round-trip cost, as well as the likelihood of having the bucket evicted
> from the cache at all.
>
> Chris and I played around with this idea by hacking the constants in
> sandboxes, verifying that larger buckets seem to help reduce the load
> times.  They should also produce fewer un-resolvable conflicts (splits),
> which should help write performance under load, too.
>
> I pushed a branch[1] today which adds that configurability as
> constructor-only parameters,, e.g.::
>
>    from BTrees.OOBTree import OOBTree
>    tree = OOBTree(max_btree_size=500, max_bucket_size=1000)


Maybe I missed it, but I doesn't look like these parameters are being
passed to internal nodes.

>
> If you don't supply the values, they fall back to the current values set
> via '#define DEFAULT_MAX_{BTREE,BUCKET}_SIZE'.
>
> Unfortunately, as we went to work on the '__getstate__' /
> '__setstate__', Chris and I realized that there was no way to make the
> persistent state backward compatible:  BTrees and TreeSets store state
> as one of:
>
> - - None (empty tree / set)
>
> - - One-element tuple (tree / set with only one bucket)
>
> - - Two-element tuple (tree / set with multiple buckets).

This is an important optimization IMO, because it allows app
developers to be able to use BTrees for small collections without
worrying about the overhead of a second persistent object.
It's not mostly a matter of pickle size, but making BTrees reasonable
to use for small collections (that might need to grow to big ones).

>
> The overhead should be pretty minimal:  most of the frequent operations
> will still be implemented in C for the BBB wrappers.
>
> Any thoughts?

I've thought about this problem for a while and had always planned to
do the configuration at the class level:

- Add the configuration as class attributes.

  This means that you can do global experiments without recompilation
  and without updating application code.

  This avoids the __setstate__ issue.

  It also avoid having to plumb arguments through to internal nodes.

- Make BTrees subclassable.

  Then apps can create new BTree types that have different size
  configurations.  If instance dictionaries are disabled for instances
  of subclasses, this is probably easy.

  (It only occurred to me just now that subclassability is probably
  easy if you disallow instance dictionaries, which I think you can do
  with a flag in a type data structure.)

The downside of this approach is that it requires a class dictionary
lookup to get the sizes.  This might be mitigated by caching the sizes
on the BTree instances.  The cache would let you reuse a lot of what
you've already done.  Basically, rather than getting the values from
an argument list, get them from the class in __init__ and __setstate__.

jimfulton:
On Thu, Mar 13, 2014 at 12:09 PM, Jim Fulton <jim@zope.com> wrote:
> On Tue, Mar 11, 2014 at 5:46 PM, Tres Seaver <tseaver@palladion.com> wrote:
>> -----BEGIN PGP SIGNED MESSAGE-----
>> Hash: SHA1
>>
>> Background
>> ==========
>>
>> In order to tune performance on large sites, I'd like to be able to
>> experiment with different bucket / tree sizes.  For instance, we have
>> sites where the number of loads for buckets is much too high for some
>> kinds of requests:  increasing the bucket size could reduce the
>> round-trip cost, as well as the likelihood of having the bucket evicted
>> from the cache at all.
>>
>> Chris and I played around with this idea by hacking the constants in
>> sandboxes, verifying that larger buckets seem to help reduce the load
>> times.  They should also produce fewer un-resolvable conflicts (splits),
>> which should help write performance under load, too.

At ZC, we've used custom builds to increase these sizes too.
We should almost certainly increase the defaults.  What do you
think, 2x? 3x?

>>
>> I pushed a branch[1] today which adds that configurability as
>> constructor-only parameters,, e.g.::

,,,

> I've thought about this problem for a while and had always planned to
> do the configuration at the class level:
>
> - Add the configuration as class attributes.
>
>   This means that you can do global experiments without recompilation
>   and without updating application code.
>
>   This avoids the __setstate__ issue.
>
>   It also avoid having to plumb arguments through to internal nodes.
>
> - Make BTrees subclassable.

Well, they are subclassable now. Someone even added the ability to
override bucket types.  Unfortunately, subclasses don't pickle correctly.
If you add instance attributes or slots, they aren't saved.

>   Then apps can create new BTree types that have different size
>   configurations.  If instance dictionaries are disabled for instances
>   of subclasses, this is probably easy.
>
>   (It only occurred to me just now that subclassability is probably
>   easy if you disallow instance dictionaries, which I think you can do
>   with a flag in a type data structure.)

I thought there was a flag, but I couldn't find it. :(

> The downside of this approach is that it requires a class dictionary
> lookup to get the sizes.  This might be mitigated by caching the sizes
> on the BTree instances.  The cache would let you reuse a lot of what
> you've already done.  Basically, rather than getting the values from
> an argument list, get them from the class in __init__ and __setstate__.

Here's a pull request:

  https://github.com/zopefoundation/BTrees/pull/8

You specify node sizes as class attributes of BTree subclasses.

It only does the lookup if necessary and caches the result
on the BTree instance.

I didn't solve the subclass pickling issue, as it was already
there.

jimfulton:
I never increased the default sizes. I think we should.

Krilivye:
Thanks @jimfulton for the details, Subclassing is a good solution!

PR #30, opened 2016-04-05
Appveyor script

fgregg:
This appveyor script will 

* built and test BTrees on windows for multiple versions of python and 32 and 64bit architectures
* on a tag, deploy window wheels to pypi

You will need to 
1. setup appveyor for the BTrees repo: https://www.appveyor.com/docs
2. create an encrypted variable for "password=the_zope.wheelbuilder_password" and let me know what it is, I'll incorporate it into the script. https://www.appveyor.com/docs/build-configuration#secure-variables

PR #33, opened 2016-04-14
Don't use/poison global wheel cache when building w/ PURE_PYTHON.

tseaver:


jamadden:
LGTM.

tseaver:
I will merge after Travis passes.  The appveyor failures are expected until we land #30.

jamadden:
Travis runs 'python setup.py test' so it isn't testing tox.ini AFAICS.

tseaver:
D'oh, should've realized.

mgedmin:
I've found a setting called "Skip branches without appveyor.yml" and enabled it.  (Ought to be default IMHO, but I can see why AppVeyor decided otherwise.)

Issue #32, opened 2016-04-14
`test_extremes` failing with 64 bit python 3 on Windows


fgregg:
https://ci.appveyor.com/project/mgedmin/btrees/build/1.0.6/job/qg5a0mkrfo00jyg2#L51
```
ERROR: test_extremes (BTrees.tests.test_OLBTree.OLBTreeTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\projects\btrees\BTrees\tests\test_OLBTree.py", line 120, in test_extremes
    btree['SMALLEST_64_BITS'] = SMALLEST_64_BITS
OverflowError: Python int too large to convert to C long
```

All 32 bit versions work fine, 64 bit python 2.7 passes as well. https://ci.appveyor.com/project/mgedmin/btrees/build/1.0.6

mgedmin:
Let's get the broken test fixed (or at least skip it on 64-bit windows with @unittest.skipIf or something) before merging this.

mgedmin:
@tseaver: you wrote that test.

My gut feeling is that this exposes a real bug in OLBTree implementation: it can hold 64-bit values on 32-bit Windows, but not on 64-bit Windows?  Most illogical!

fgregg:
This test is skipped for 32 bit architecture https://github.com/zopefoundation/BTrees/blob/master/BTrees/tests/test_OLBTree.py#L112

tseaver:
I believe this error is due to a Windows oddity: [`long` is 32-bit on 64-bit Windows](https://msdn.microsoft.com/en-us/library/3b2e7499.aspx).

tseaver:
Which is maybe the stupidest design choice I can think of:  likely due to some crazy notion of preserving ABI with Windows 95. :(

fgregg:
That is weird!

mgedmin:
Shall we declare Win64 to be a 32-bit architecture and skip this test too?

tseaver:
What does `platform.architecture()` return on Win64? [1]

[1] https://docs.python.org/2/library/platform.html#platform.architecture

Maybe we should be using the workaround from that API reference:

```python
def _skip_on_32_bits(wrapped):
    if sys.maxsize == 2**32:
        def _dummy(*args):
            pass
        return _dummy
    return test_method
```

mgedmin:
+1 for the sys.maxsize check

tseaver:
LOL

tseaver:
[...]

PR #34, opened 2016-04-14
Use 'sys.maxsize' to detect real 32-bit Python.

tseaver:
Works around the fact that `platform.architecture()` reports size of *pointers*, rather than size of *longs*, which matters on Win64.

Closes: #32.


tseaver:
Closing in favor of #30, where the Win64 bits will actually be tested.

PR #38, opened 2016-04-15
Use unittest.skip to skip tests

mgedmin:
This way skipped tests will be visible in the test runner output.


jamadden:
242 tests skipped! Yay! Wait, no...

Travis is green, LGTM (assuming AppVeyor gets around to building them and they'll be green there too.)

PR #43, opened 2016-08-03
More consistency between Python and C impls.

### LINKED ISSUE #42: BTree.get(object()) raises TypeError on CPython, returns default on PyPy
I ran into an implementation difference between the C version and the Python versions.

Here's PyPy:
```python
Python 2.7.10 (7e8df3df9641, Jun 14 2016, 13:30:54)
[PyPy 5.3.1 with GCC 4.2.1 Compatible Apple LLVM 5.1 (clang-503.0.40)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>>> import BTrees
>>>> BTrees.OOBTree.OOBTree().get(object())
>>>>
```

Here's CPython:
```python
Python 2.7.12 (default, Jul 11 2016, 16:16:26)
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import BTrees
>>> BTrees.OOBTree.OOBTree().get(object())
>>>
```

The specific type of object with default comparison doesn't really matter. And they both raise a TypeError if you try to set a key with default comparison (albeit with different error messages).

Which is the right behaviour here? I would argue that the Python behaviour is certainly nicer and friendlier to the user. It makes it easier to substitute a BTree for a dict in more places. In fact, on CPython, I had to subclass the BTree to be able to use it in place of a dict because of this TypeError (even though I had already made sure that we would never try to set such a value, sometimes we still query for them).

If there's consensus I can put together a PR to bring them into line.

jamadden:
Fixes #42.

See CHANGES.rst for details on what changes. Additional test cases cover
the changed behaviour.

### DIFF
--- a/BTrees/BTreeTemplate.c
+++ b/BTrees/BTreeTemplate.c
@@ -219,6 +219,8 @@ BTree_check(BTree *self)
     return result;
 }
 
+#define _GET_KEY_ERROR 1
+#define _GET_ALL_ERROR 0
 /*
 ** _BTree_get
 **
@@ -229,6 +231,9 @@ BTree_check(BTree *self)
 **      keyarg      the key to search for, as a Python object
 **      has_key     true/false; when false, try to return the associated
 **                  value; when true, return a boolean
+**      ignore_key_err    true/false: When true, ignore the TypeError from
+**                         a key conversion issue, and return NULL with
+**                         a KeyError set.
 ** Return
 **      When has_key false:
 **          If key exists, its associated value.
@@ -239,14 +244,22 @@ BTree_check(BTree *self)
 **          If key doesn't exist, 0.
 */
 static PyObject *
-_BTree_get(BTree *self, PyObject *keyarg, int has_key)
+_BTree_get(BTree *self, PyObject *keyarg, int has_key, int ignore_key_err)
 {
     KEY_TYPE key;
     PyObject *result = NULL;    /* guilty until proved innocent */
     int copied = 1;
 
     COPY_KEY_FROM_ARG(key, keyarg, copied);
-    UNLESS (copied) return NULL;
+    UNLESS (copied)
+    {
+        if (ignore_key_err && PyErr_ExceptionMatches(PyExc_TypeError))
+        {
+            PyErr_Clear();
+            PyErr_SetObject(PyExc_KeyError, keyarg);
+        }
+        return NULL;
+    }
 
     PER_USE_OR_RETURN(self, NULL);
     if (self->len == 0)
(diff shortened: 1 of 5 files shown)


tseaver:
I restarted the `homebrew2` build, which was terminated at 10 minutes.

BTrees/tests/test_OOBTree.py
```diff
@@ -126,6 +126,12 @@ class C(object):
 
         self.assertRaises(TypeError, lambda : t.__setitem__(C(), 1))
 
+        try:
+            t[C()] = 1
+            self.fail("Should raise TypeError")
+        except TypeError as e:
+            self.assertEqual(e.args[0], "Object has default comparison")
+
```
tseaver: ISTM this would be clearer / more idiomatic:

```python
        with self.assertRaises(TypeError) as raises:
            t[C()] = 1

        self.assertEquals(raiess.exception.args[0], "Object has default comparison")
```

See: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises

BTrees/BTreeTemplate.c
```diff
@@ -219,6 +219,8 @@ BTree_check(BTree *self)
     return result;
 }
 
+#define _GET_KEY_ERROR 1
+#define _GET_ALL_ERROR 0
```
tseaver: In conjunction with the `ignore_key_err` parameter name below, these names seem a little weird to me:  could we come up with clearer ones?

jamadden:
Thank you. I have no idea why it hung. The restart seems to have taken care of it.

BTrees/tests/test_OOBTree.py
jamadden: I always forget that it can be used as a context manager. Thanks, will do!
(I write most of my tests with PyHamcrest, where you would write
```python
assert_that( calling(t.__setitem__).with_args(C(), 1),
                     raises(TypeError, "Object has default comparison"))
```
)

BTrees/BTreeTemplate.c
jamadden: Sure. I'm open to suggestions!

My process here was:
- The param was originally named something like `ignore_def_cmp`, before I realized that the int-trees *also* raised a TypeError, just not the default-comparison type error.
- I had updated all the `_BTree_get` calls to add the param. But having two "boolean" params at the end of the call, so it looked like `_BTreeGet(self, key, 0, 1)` was pretty unreadable.
- Thinking about the callers, they either want `_BTree_get` to raise *all* lookup-related errors, or just a KeyError. Hence the names of the params, so now we have `_BTree_get(self, key, 0, _GET_KEY_ERROR)`
- Then I realized the thing about integer trees and changed the param name...which I'm just now realizing is a bit confusing because we're ignoring a TypeError *on* the key...

tseaver:
LGTM

jamadden:
Thank you for the review. I will merge soon if there are no other comments. (I still had a hard time with the naming of the C params/macros, even after your help, which suggests there's something missing, but I don't know what it is.)

PR #45, opened 2016-09-29
Add requirements file for RTD.

jamadden:
This allows BTrees docs to build and be published on http://btrees.readthedocs.io. This is useful for intersphinx references. (The docs on pythonhosted.org/BTrees haven't been built since 2012 or so and don't have intersphinx data.)


jamadden:
Thanks! Sadly, there is [not yet](https://github.com/pypa/pypi-legacy/pull/516) a way to remove the automatically-generated link to the pythonhosted.org docs, but we can at least update the README.rst to point to the new site.

PR #47, opened 2016-10-28
BTreePy: Reset ._data unconditionally in .clear()

navytux:
_Tree.clear() is called by _Tree.__setstate__() to reset both .data and
._firstbucket before actually proceeding with setting internal state
with data loaded from database.

However since 0718a45e (Use a Python **reduce** method to make pickles
match C.) ._data reset became conditional, first checking that ._data is
not already empty.

The .**setstate**() is often called with object being a ghost.  However
ZODB was never promising to keep attributes alive on ghost objects and
in fact attributes kept in **dict** was always released, but attributes
in slotted variables, probably due to overlook, started to be released
on ghostification only recently:

```
https://github.com/zopefoundation/persistent/commit/fe2219f4
https://github.com/zopefoundation/persistent/pull/44
```

and that change causes many BTree tests to fail in pure-python part, e.g.:

```
ERROR: testGhostUnghost (BTrees.tests.test_IIBTree.IIBTreeTest)
  File "BTrees/tests/common.py", line 122, in testGhostUnghost
    self.assertEqual(list(root2[i].items()) , list(t.items()))
  File "/home/kirr/src/wendelin/z/persistent/persistent/persistence.py", line 266, in __getattribute__
    oga(self, '_p_activate')()
  File "/home/kirr/src/wendelin/z/persistent/persistent/persistence.py", line 386, in _p_activate
    jar.setstate(self)
  File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/Connection.py", line 771, in setstate
    self._reader.setGhostState(obj, p)
  File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/serialize.py", line 623, in setGhostState
    obj.__setstate__(state)
  File "BTrees/_base.py", line 1069, in __setstate__
    self.clear()
  File "BTrees/_base.py", line 818, in clear
    if self._data:
  File "/home/kirr/src/wendelin/z/persistent/persistent/persistence.py", line 268, in __getattribute__
    return oga(self, name)
AttributeError: _data
```

Via unconditionally resetting ._data in clear we can make the problem go away.

After this patch with PURE_PYTHON=1 BTree testsuite is passing fully again.

/cc @jamadden

### DIFF
--- a/BTrees/_base.py
+++ b/BTrees/_base.py
@@ -815,9 +815,7 @@ class _Tree(_Base):
         self._del(self._to_key(key))
 
     def clear(self):
-        if self._data:
-            # In the case of __init__, this was already set by __new__
-            self._data = []
+        self._data = []
         self._firstbucket = None
 
     def __nonzero__(self):


navytux:
About CI being red: it is only somtimes infrastructure problems not related to the patch, like:

- "Error: python-2.7.12_1 already installed"  (e.g. https://travis-ci.org/zopefoundation/BTrees/jobs/171340652), or
- "/Users/travis/build.sh: line 109: shell_session_update: command not found" (e.g. https://travis-ci.org/zopefoundation/BTrees/jobs/171340653)

besides that all tests are passing.

BTrees/_base.py
```diff
@@ -815,9 +815,7 @@ def __delitem__(self, key):
         self._del(self._to_key(key))
 
     def clear(self):
-        if self._data:
-            # In the case of __init__, this was already set by __new__
-            self._data = []
+        self._data = []
```
jamadden: What about `if not hasattr(self, '_data') or self._data:` instead of the unconditional set, and extending the comment to mention that `_data` can sometimes be blown away entirely when the object is ghosted?  That seems more explicit to me, and has the minor benefit of avoiding object churn.

jamadden: Maybe what I really don't like is how `clear` is called when we're in an entirely unknown state. Imagine a subclass of the BTree (ill-advised, but known to happen) that does something interesting in its clear method---maybe records some stats to graphite or something. If `clear` is called by `__setstate__` and during steady state, that subclass will now be broken, when it wasn't before.

- It's not safe for `__setstate__` to call arbitrary methods until after it's done.
- `__setstate__` is already coupled to `__getstate__` and the internal implementation of the object.

Maybe it would be better for `__setstate__` to simply assign to `self._data` *itself* and avoid the call to clear entirely. Then there's no object churn for the common case of `__init__` -> `clear`, and there's no implicit assumptions about the object state in `__setstate__`, it's all explicit. 

I think I like that best of all, so far.

navytux: I'm not agains for having `__setstate__` setting `._data` itself directly, but the code outline with `clear()` was already there and I've tried to only fix things up. Regarding pure-python codepaths my approach is only try to not to break it because here at Nexedi we only use original C-based codes.

In any case, be it separate `.clear()` or setting `._data` directly in `__setstate__()`, I would uncoditinally initialize `._data` to [] until there is a clear performance win in not doing so.

> If clear is called by __setstate__ and during steady state, that subclass will now be broken, when it wasn't before.

Could you please clarify? current state is `__setstate__` is already calling `clear`. Why it was not broken before then?

jamadden: > Could you please clarify? current state is __setstate__ is already calling clear. Why it was not broken before then?

For exactly the same reason BTrees in general broke.

navytux: I see, thanks. So it is not clear-specific but just a general things about variables in slots. My approach, once again, would be to fix those cases one-by-one with, as you suggested, bumping persistent version for the slots ghostification patch.

It would be a pity not to be able to rely on ghostification semantics to completely release objects.

jamadden: > I see, thanks. So it is not clear-specific but just a general things about variables in slots. My approach, once again, would be to fix those cases one-by-one with, as you suggested, bumping persistent version for the slots ghostification patch.

Yes, it is `clear` specific. `clear` is (now) being called when the object's state is undefined. The state that's was supposed to be an invariant, established by `__new__` and maintained by `__setstate__`. That invariant is now broken, and calling a method when the invariants of the class do not hold is a sure way to have trouble (not to mention just being rude). Since the invariant is now broken, `__setstate__` should not call into code that *may or may not* rely on that invariant---it has no way of knowing.

It is far better to avoid introducing trouble at the chokepoint, where it can be fixed all at once---the superclass--than expect everyone else to go clean up when their code breaks.

> It would be a pity not to be able to rely on ghostification semantics to completely release objects.

First, we've never been able to rely on that. persistent has never done that. This change obviously breaks a number of cases. We've located two just by looking at a handful of (very widely used) projects. 

It starting to seem to me that changing this and breaking unknown downstream projects is far more of a risk than simply using `__dict__` for objects where allowing `__slots__` to remain is a problem. 

Second, it's *still* not a guarantee. When the object is implemented in C, if it doesn't specifically do something with its struct members, they aren't released. This is the only reason that zope.container's Proxy still works in the C implementation after this change---the struct member that is `_wrapped` is left sitting untouched by `_p_deactivate`.

navytux: I see. I hope my idea about `__keep__` in https://github.com/zopefoundation/persistent/pull/44#issuecomment-256998968 should allow retaining working state and compatibility more or less easily.

I mean for BTree case we could just first add `__keep__ = ['_data'] # or __slots__` and then it works as it was working before. Then we can remove the `__keep__` alongside reworking `__setstate__` to not call clear but directly setting `._data`

I hope this way it could be made that everyone are happy.

navytux:
Update:

1. for backward-compatibility reason it was decided to release `__slots__` automatically only for classes that override `__new__`:
  https://github.com/zopefoundation/persistent/pull/44#issuecomment-261019084  
  and the patch that do that compatibility fixup for persistent is there:
  https://github.com/zopefoundation/persistent/pull/50

2. `BTrees._base._Tree` overrides `__new__`.

Thus with fixed persistent (https://github.com/zopefoundation/persistent/pull/50) all BTrees tests pass in pure-python mode out of the box without any patch to BTrees.

This way here present patch can be completely droppped (or reworked by pure-python mode maintainers to their taste if they wish so).

jamadden:
> for backward-compatibility reason it was decided to release __slots__ automatically only for classes that override __new__:
zopefoundation/persistent#44 (comment)

Typo: For classes that **don't** override `__new__`. If it was the other way around, *then* BTrees would be broken.

So yes, given that update, I agree, this can be closed.

navytux:
> Typo

Oops, thanks for catching and correcting it :)  Of course I've meant classes that don't override `__new__`.

Closing.

Issue #48, opened 2016-11-06
BTrees setup.py doesn't respect user defined index-url setting, breaking pip and requirements.txt style installs


agreene5:
When installing BTrees, the index-url of all dependencies seem to be "forced" to pypi.python.org regardless of the index-url setting in requirements.txt or in ~/.pip.ini

This creates serious problems in environments where Artifactory is used as a "proxy" to pypi.python.org and where direct connectivity to pypi.python.org is not permitted by policy. This causes the install to hang, and then finally fail with an error saying it failed to retrieve a pypi.python.org URL.

This is easy to reproduce if you have an Artifactory server or some other non pypi.python.org PyPi repository URL configured in your ~/.pip.ini like this:

--- snip ---
[global]
index-url = http://my.local.repo/artifactory/api/pypi/pypi-repos/simple
--- snip ---

When I do pip install BTrees, the first HTTP/HTTPS request is made to grab the BTrees package from my local Artifactory server, as expected. After that, the 'persistent' package dependency is fetched since it is a dependency of BTrees- but it is not fetched from the Artifactory server as one would expect- instead, it always goes directly to pypi.python.org to try to fetch the package.

The workaround is to manually install all of the dependencies before install BTrees, so that it doesn't need to fetch any dependencies.

I would like to submit a pull request to fix the issue, but I'm not terribly familiar with setuptools or pip. Maybe this is an easy fix for someone who knows the tools better?

Thanks.

agreene5:
The issue seems to be this:

https://bitbucket.org/tarek/distribute/issues/338/unable-to-specify-custom-index-for

agreene5:
Submitted #49 for this - the conclusion everyone seems to come to is "don't use setup_requires" - and it doesn't seem to be needed here

agreene5:
For now, I will stick with a client-side workaround. I am using the following to get ZEO+ZEODB working via a single run with requirements.txt:

$ cat venv/.pydistutils.cfg 
[easy_install]
allow_hosts = *.internal
find_links = 
    http://my.artifactory.internal/artifactory/api/pypi/pypi-repos/simple/zope.interface/
    http://my.artifactory.internal/artifactory/api/pypi/pypi-repos/simple/persistent/

And the requirements.txt file I am using:

$ cat venv/requirements.txt
BTrees==4.3.1
ZEO==5.0.2
ZODB==5.0.0

This solves the issue. Having to have specific find_links values is a little annoying, but isn't terrible.

Issue #4, opened 2014-03-15
clang error on install (Mac OSX)


cs0sor:
Getting a clang error when trying to install ZODB. Some issue with compiler arguments. Have latest Command Line Tools and Xcode 5.1

creating build/temp.macosx-10.9-intel-2.7/src/BTrees

cc -fno-strict-aliasing -fno-common -dynamic -arch x86_64 -arch i386 -g -Os -pipe -fno-common -fno-strict-aliasing -fwrapv -mno-fused-madd -DENABLE_DTRACE -DMACOSX -DNDEBUG -Wall -Wstrict-prototypes -Wshorten-64-to-32 -DNDEBUG -g -fwrapv -Os -Wall -Wstrict-prototypes -DENABLE_DTRACE -arch x86_64 -arch i386 -pipe -Isrc -I/System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7 -c src/BTrees/_OOBTree.c -o build/temp.macosx-10.9-intel-2.7/src/BTrees/_OOBTree.o

clang: error: unknown argument: '-mno-fused-madd' [-Wunused-command-line-argument-hard-error-in-future]

clang: note: this will be a hard error (cannot be downgraded to a warning) in the future

error: command 'cc' failed with exit status 1

Clang + LLVM Version:

> gcc --version
> Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
> Apple LLVM version 5.1 (clang-503.0.38) (based on LLVM 3.4svn)
> Target: x86_64-apple-darwin13.1.0
> Thread model: posix

PR #49, opened 2016-11-07
Proposed fix for issue #4

agreene5:
Removed setup_requires = ['persistent']

The issue: #48 

Since the persistent module is not actually needed at setup time- it is only needed in install_requires, where it is already specified.

When it is present in setup_requires, it breaks configurations that use local repositories
like Artifactory. There are no clean workarounds that I have been able to find.

Since the persistent module is not actually needed at setup time- it is only needed in install_requires, where it is already specified.

When it is present in setup_requires, it breaks configurations that use local repositories
like Artifactory. There are no clean workarounds that I have been able to find.

I tested this and it works fine. Perhaps this was done by mistake?

I'd appreciate if this could be merged into a 4.3.0a or 4.3.1 release on PyPi so I can actually benefit from this being fixed. As of now I am left with clunky workarounds and can't use requirements.txt
I tested this and it works fine. Perhaps this was done by mistake?

I'd appreciate if this could be merged into a 4.3.0a or 4.3.1 release on PyPi so I can actually benefit from this being fixed. As of now I am left with clunky workarounds and can't use requirements.txt

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,7 +144,6 @@ setup(name='BTrees',
       include_package_data=True,
       zip_safe=False,
       ext_modules = ext_modules,
-      setup_requires=['persistent'],
       extras_require = {
         'test': TESTS_REQUIRE,
         'ZODB': ['ZODB'],


tseaver:
@agreene5 `persistent` is a *build-time* requirement:  it provides the header files and libraries needed to link the `BTrees` C extensions.

agreene5:
ah, my mistake. I just removed the line and it seemed to work fine via python setup.py install. Perhaps I missed something (seems likely)

I found a few workarounds via pydistutils.cfg but they were one off / hack type solutions. I guess this PR can be closed and I'll go back to the drawing board. If you have any ideas on how I can cleanly work around this, comment in the issue I entered. If you don't think there's anything that can on the BTrees side, you can close the issue as well. I just can't imagine there not being a way to do this cleanly..hrmm

Issue #51, opened 2016-11-20
C BTrees should allow unpickling with None/unorderable keys


jamadden:
The Python implementation currently allows this.

See https://groups.google.com/d/msg/zodb/xhVM0ejl6aE/7mXRaZeEAQAJ

Issue #53, opened 2016-11-20
C BTrees should allow getting/deleting unorderable keys


jamadden:
Keys such as None currently result in KeyError in the subscript cases, even if they're in the tree. (How are they in the tree? Unpickling old data, or #52.) We should be able to delete them.

The Python implementation allows this.

See https://groups.google.com/d/msg/zodb/xhVM0ejl6aE/NMBsQDNcAQAJ

PR #54, opened 2016-11-27
Check for object key default comparison on set only

davisagli:
This makes the C implementation compatible with the Python implementation by only checking keys for default comparison when they are being set. (see #51)

It also makes it possible to delete unorderable keys, on Python 2 only. (see #53)

### DIFF
--- a/BTrees/BTreeTemplate.c
+++ b/BTrees/BTreeTemplate.c
@@ -726,6 +726,10 @@ _BTree_set(BTree *self, PyObject *keyarg, PyObject *value,
     COPY_KEY_FROM_ARG(key, keyarg, copied);
     if (!copied)
         return -1;
+#ifdef KEY_CHECK_ON_SET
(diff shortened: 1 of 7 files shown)


tseaver:
LGTM

jamadden:
@davisagli Would you like to merge this? Then I believe we can close #51 and #53.  Well actually, I think I want to come back and add an explicit pickle-based test for #51 just to be safe.

davisagli:
@jamadden I'd be happy to see it merged. I wasn't sure if you or @jimfulton had reviewed it yet, and I thought I wasn't supposed to merge my own pull request.

jamadden:
@tseaver reviewed it and that's good enough for me 😄 But it also LGTM. I think when the PR submitter is also a committer, once @tseaver says "LGTM" it's usual that the PR submitter will do the final merge---at least, that's the workflow I've been assuming 😉

davisagli:
Okay, fine with me.

PR #55, opened 2016-12-22
Get 'py35-pure' tox environment passing.

tseaver:
On Py3k, the conditional imports fail with an `AttributeError` when the C extensions are missing, rather than an `ImportError`.

### DIFF
--- a/BTrees/IFBTree.py
+++ b/BTrees/IFBTree.py
@@ -106,7 +106,7 @@ weightedIntersectionPy = _set_operation(_weightedIntersection, IFSetPy)
 
 try:
     from ._IFBTree import IFBucket
-except ImportError: #pragma NO COVER w/ C extensions
+except (ImportError, AttributeError): #pragma NO COVER w/ C extensions
     IFBucket = IFBucketPy
     IFSet = IFSetPy
     IFBTree = IFBTreePy
--- a/BTrees/IIBTree.py
+++ b/BTrees/IIBTree.py
@@ -107,7 +107,7 @@ weightedIntersectionPy = _set_operation(_weightedIntersection, IISetPy)
 
 try:
     from ._IIBTree import IIBucket
-except ImportError: #pragma NO COVER w/ C extensions
+except (ImportError, AttributeError): #pragma NO COVER w/ C extensions
     IIBucket = IIBucketPy
     IISet = IISetPy
     IIBTree = IIBTreePy
--- a/BTrees/IOBTree.py
+++ b/BTrees/IOBTree.py
@@ -89,7 +89,7 @@ multiunionPy = _set_operation(_multiunion, IOSetPy)
 
 try:
     from ._IOBTree import IOBucket
-except ImportError: #pragma NO COVER w/ C extensions
+except (ImportError, AttributeError): #pragma NO COVER w/ C extensions
     IOBucket = IOBucketPy
     IOSet = IOSetPy
     IOBTree = IOBTreePy
--- a/BTrees/LFBTree.py
+++ b/BTrees/LFBTree.py
@@ -107,7 +107,7 @@ weightedIntersectionPy = _set_operation(_weightedIntersection, LFSetPy)
 
 try:
     from ._LFBTree import LFBucket
-except ImportError: #pragma NO COVER w/ C extensions
+except (ImportError, AttributeError): #pragma NO COVER w/ C extensions
     LFBucket = LFBucketPy
     LFSet = LFSetPy
     LFBTree = LFBTreePy
--- a/BTrees/LLBTree.py
+++ b/BTrees/LLBTree.py
@@ -107,7 +107,7 @@ weightedIntersectionPy = _set_operation(_weightedIntersection, LLSetPy)
 
 try:
     from ._LLBTree import LLBucket
-except ImportError: #pragma NO COVER w/ C extensions
+except (ImportError, AttributeError): #pragma NO COVER w/ C extensions
     LLBucket = LLBucketPy
     LLSet = LLSetPy
     LLBTree = LLBTreePy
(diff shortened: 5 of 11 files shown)


tseaver:
The failures only showed up on [jenkins.pylonsproject.org](http://jenkins.pylonsproject.org/job/BTrees/88/console):  Travis / Appveyor don't run the `PURE_PYTHON` environments.

jamadden:
For ease of reference, here's the error:
```
  File "/home/agendaless/jenkins/pylonsproject/jobs/BTrees/workspace/BTrees/_base.py", line 33, in <module>
    class _Base(Persistent):
  File "/home/agendaless/jenkins/pylonsproject/jobs/BTrees/workspace/BTrees/_base.py", line 51, in _Base
    __import__('BTrees._OOBTree')
AttributeError: module 'persistent' has no attribute 'cPersistence'
```

That seems very weird to me. Do you know what's actually causing that? (Is persistent PURE_PYTHON tested on 3.5 anywhere, it seems like it could be vulnerable too.) This seems more like a Python bug to me. 

The closest mention of anything I could find in the 3.5 release notes was - "[Issue #24492](http://bugs.python.org/issue24492): A "package" lacking a __name__ attribute when trying to perform a ``from .. import ...`` statement will trigger an ImportError instead of an AttributeError;" there is also at least one older release note about `__import__` failing to propagate an ImportError in 3.4, which seems suspiciously similar to what may be going on here ("- Issue #15111: __import__ should propagate ImportError when raised as a  side-effect of a module triggered from using fromlist.")

tseaver:
@jamadden I just added https://github.com/zopefoundation/persistent/pull/56, which didn't have any issues.

jamadden:
FWIW, I can't reproduce this in a fresh virtualenv and checkout with Python 3.5.2 on OS X:
```
$ mktmpenv -p /opt/local/bin/python3.5
...
$ pip install tox
$ hub clone zopefoundation/BTrees
$ cd BTrees
$ tox -e py35-pure
GLOB sdist-make: /Users/jmadden/Projects/VirtualEnvs/tmp-9b1564a8400cc00a/BTrees/setup.py
py35-pure create: /Users/jmadden/Projects/VirtualEnvs/tmp-9b1564a8400cc00a/BTrees/.tox/py35-pure
py35-pure installdeps: zope.interface, persistent, transaction
py35-pure inst: /Users/jmadden/Projects/VirtualEnvs/tmp-9b1564a8400cc00a/BTrees/.tox/dist/BTrees-4.3.2.dev0.zip
py35-pure installed: BTrees==4.3.2.dev0,persistent==4.2.2,transaction==2.0.3,zope.interface==4.3.3
py35-pure runtests: PYTHONHASHSEED='247444608'
py35-pure runtests: commands[0] | python setup.py -q test -q
...
$ ls -l .tox/py35-pure/lib/python3.5/site-packages/*so
gls: cannot access '/Users/jmadden/Projects/VirtualEnvs/tmp-9b1564a8400cc00a/BTrees/.tox/py35-pure/lib/python3.5/site-packages/*so': No such file or directory
```

jamadden:
Any chance the Jenkins test directory wasn't clean? Because 'BTrees' is a directory in the tox default working directory, anything in it can be imported. So if there was a built C extension there, it would be imported even in PURE_PYTHON mode by the code in _base.py.  I've run into this causing problems on other projects that skip having a `src/` directory for their package (notably relstorage)...that's actually why I went with the fresh checkout and working dir in the above example, because the first time I tried to test py35-pure it turned out I had already had built C extensions that were getting picked up.

tseaver:
I just [cleared the workspace and reran on Jenkins](http://jenkins.pylonsproject.org/job/BTrees/89/console) and they still fail.

jamadden:
Any idea what version of Python 3.5 it has?

jamadden:
I also can't reproduce this on Ubuntu 16.04 with Python 3.5.2 in a fresh checkout/venv.

tseaver:
@jamadden Try running `tox -e py35,py35-pure` in your fresh checkout.  :(

jamadden:
Ok, I tried that. I still didn't have any failure on OS X, but I did on Ubuntu. It seems like tox is leaving the compiled .so files around in the main BTrees directory...and since that's on the sys.path as described above (because it's the working directory) we get this problem. The `cPersistence` module wasn't even compiled because of PURE_PYTHON, but the C BTree modules were, *and* the BTree modules don't respect the runtime PURE_PYTHON variable like persistent does. So we attempt to load the C BTree module, which uses the Capsule API to load `persistent.cPersistence.CAPI`...and I'm guessing it just takes the AttributeError exception and doesn't alter it into an ImportError.

So maybe a better fix would be to have the BTreeModuleTemplate.c look for the NULL return and raise an ImportError explicitly? That would seem to fix the root cause.

(I may have some of the exact details wrong, this is my first time looking at the way those modules interact at the C level.)

Or we could just change the tox working directory to fix the test and say that having BTree C code but no persistent C code (or vice versa) just isn't supported...which, it really isn't. As it stands, if the PURE_PYTHON variable is set, but the extensions are both available, Python code that subclasses Persistent is going to get a different thing than the BTree code is doing.

jamadden:
This patch to BTreeModuleTemplate seems to make the `tox -e py35,py35-pure` case work on linux for me:
```patch
@@ -558,8 +558,12 @@ module_init(void)
     cPersistenceCAPI = (cPersistenceCAPIstruct *)PyCObject_Import(
                 "persistent.cPersistence", "CAPI");
 #endif
-    if (cPersistenceCAPI == NULL)
+    if (cPersistenceCAPI == NULL) {
+       if (PyErr_Occurred() && !PyErr_ExceptionMatches(PyExc_ImportError)) {
+           PyErr_SetString(PyExc_ImportError, "<No CPersistence>");
+       }
         return NULL;
+   }

 #ifdef PY3K
 #define _SET_TYPE(typ) ((PyObject*)(&typ))->ob_type = &PyType_Type
```

jamadden:
Closed by #57

PR #56, opened 2016-12-22
Python/C OO consistency

### LINKED ISSUE #50: C/Py difference: keys that are the same object
The C implementation does a comparison [by first checking whether the two keys are the same pointer](https://github.com/python/cpython/blob/2.7/Objects/object.c#L997); the Python implementation just goes [right to the `==` operator](https://github.com/zopefoundation/BTrees/blob/master/BTrees/_base.py#L114). In some cases of (broken?) objects this leads to a discrepancy: The C implementation can find a key, but the Python implementation cannot.

I would suggest that the Python implementation should use `k is key or k == key` to clear this up.

See https://groups.google.com/forum/#!topic/zodb/xhVM0ejl6aE

### LINKED ISSUE #51: C BTrees should allow unpickling with None/unorderable keys

### LINKED ISSUE #53: C BTrees should allow getting/deleting unorderable keys

jamadden:
Check key identity before ==. This matches the C implementation.
Fixes #50

Add a pickle test to the work done in #54. Fixes #51. Fixes #53 (for
Python 2).

### DIFF
--- a/BTrees/tests/test_OOBTree.py
+++ b/BTrees/tests/test_OOBTree.py
@@ -161,12 +161,12 @@ class OOBTreeTest(BTreeTests, unittest.TestCase):
         self.assertRaises(KeyError, t.__getitem__, C())
         self.assertFalse(C() in t)
 
-    # Check that a None key can be deleted in Python 2.
-    # This doesn't work on Python 3 because None is unorderable,
-    # so the tree can't be searched. But None also can't be inserted,
-    # and we don't support migrating Python 2 databases to Python 3.
     @_skip_under_Py3k
     def testDeleteNoneKey(self):
+        # Check that a None key can be deleted in Python 2.
+        # This doesn't work on Python 3 because None is unorderable,
+        # so the tree can't be searched. But None also can't be inserted,
+        # and we don't support migrating Python 2 databases to Python 3.
         t = self._makeOne()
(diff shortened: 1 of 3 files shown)


BTrees/tests/test_OOBTree.py
```diff
@@ -352,3 +391,6 @@ def test_suite():
         unittest.makeSuite(OOTreeSetPyConflictTests),
         unittest.makeSuite(OOModuleTest),
     ))
+
+if __name__ == '__main__':
+    unittest.main(defaultTest='test_suite')
```
tseaver: While I sympathize with the motivation, we really need to avoid adding scripty baggage to the unit test modules.

jamadden: My mistake, sorry.

jamadden:
Thanks!

PR #58, opened 2017-01-10
Allow None as a special key (sorted smaller than all others).

jimfulton:
The C changes took a few minutes.

The pure python implementation took a *lot* more care (and wtf the tests?).

### DIFF
--- a/BTrees/_base.py
+++ b/BTrees/_base.py
@@ -113,7 +113,7 @@ class _BucketBase(_Base):
             k = keys[i]
             if k is key or k == key:
                 return i
-            if k < key:
+            if k is None or (key is not None and k < key):
                 low = i + 1
             else:
                 high = i
@@ -256,7 +256,7 @@ _object_lt = getattr(object, '__lt__', _marker)
 def _no_default_comparison(key):
     # Enforce test that key has non-default comparison.
     if key is None:
-        raise TypeError("Can't use None as a key")
+        return
     if type(key) is object:
         raise TypeError("Can't use object() as keys")
     lt = getattr(key, '__lt__', None)
@@ -919,7 +919,7 @@ class _Tree(_Base):
 
         max = self._to_key(max)
         index = self._search(max)
-        if index and data[index].child.minKey() > max:
+        if index and cmp(data[index].child.minKey(), max) > 0:
             index -= 1 #pragma: no cover  no idea how to provoke this
         return data[index].child.maxKey(max)
 
@@ -1014,7 +1014,7 @@ class _Tree(_Base):
             self._p_changed = True
 
         # fix up the node key, but not for the 0'th one.
-        if index > 0 and child.size and key == data[index].key:
+        if index > 0 and child.size and cmp(key, data[index].key) == 0:
             self._p_changed = True
             data[index].key = child.minKey()
 
@@ -1269,7 +1269,7 @@ class Tree(_Tree):
 
     def byValue(self, min):
         return reversed(
-                sorted((v, k) for (k, v) in self.iteritems() if v >= min))
+                sorted(((v, k) for (k, v) in self.iteritems() if v >= min)))
 
     def insert(self, key, value):
         return bool(self._set(key, value, True)[0])
--- a/BTrees/_base.py
+++ b/BTrees/_base.py
@@ -22,7 +22,7 @@ from persistent import Persistent
 
 from .Interfaces import BTreesConflictError
 from ._compat import PY3
-from ._compat import cmp
+from ._compat import compare
 from ._compat import int_types
 from ._compat import xrange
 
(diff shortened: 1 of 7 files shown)


BTrees/_compat.py
```diff
@@ -23,7 +24,17 @@
 
     int_types = int, long
     xrange = xrange
-    cmp = cmp
+    cmp_ = cmp
+    def cmp(x, y):
```
davisagli: Maybe we should call this (and the one below) something else to emphasize that it's not the standard Python `cmp`?

BTrees/_compat.py
```diff
@@ -13,6 +13,7 @@
 ##############################################################################
 import sys
 
+None_ = None
```
davisagli: What's the purpose of this alias?

BTrees/check.py
```diff
@@ -346,13 +348,13 @@ def visit_bucket(self, obj, path, parent, is_mapping,
     def check_sorted(self, obj, path, keys, lo, hi):
         i, n = 0, len(keys)
         for x in keys:
-            if lo is not None and not lo <= x:
+            if lo is not None and not cmp(lo, x) <= 0:
```
davisagli: Maybe worth a comment that `None` is treated as a sentinel in `lo` and `hi` ("without bound") rather than a literal `None`.

BTrees/_base.py
```diff
@@ -1269,7 +1269,7 @@ def iteritems(self, min=_marker, max=_marker,
 
     def byValue(self, min):
         return reversed(
-                sorted((v, k) for (k, v) in self.iteritems() if v >= min))
+                sorted(((v, k) for (k, v) in self.iteritems() if v >= min)))
```
davisagli: I don't see a substantive change here, just a spurious set of parens added.

BTrees/_base.py
```diff
@@ -113,7 +113,7 @@ def _search(self, key):
             k = keys[i]
             if k is key or k == key:
                 return i
-            if k < key:
+            if k is None or (key is not None and k < key):
                 low = i + 1
```
davisagli: I think the `key is not None` part of this is an unnecessary extra comparison, since `or` short-circuits.

BTrees/_base.py
davisagli: Do we need to use `cmp` here?

BTrees/_compat.py
jimfulton: Hm, no good purpose. I'll remove it.

BTrees/_compat.py
jimfulton: Good idea. I'll change it.

BTrees/check.py
jimfulton: Hm, good point. (This change was made fairly mechanically.)

BTrees/_base.py
jimfulton: Actually, the test above is wrong. :(

If both k and key are None, then k < key is false, which the expression above doesn't capture.

Let see:

     if key is not None and (k is None or k < key):

I should just change this to use cmp (or whatever we rename it to).

BTrees/_base.py
jimfulton: I thought we needed ``cmp``, but this is sorting *values* not keys. When I changed it back, I left the extra parents.

I'll revert this.

jimfulton: I think I was wrong about being wrong. :)

Which raises an issue.

Given that we often use reverse indexes and even given this code.....

On Python 3, a None value will error here.  I'm thinking we *should* use ``compare`` (previously ``cmp``) here.

But the inconsistency with regular Python is a little troubling.

I'm tempted to:

a) Lobby for None to be less than everything else in Python, or:

b) Create a null singleton (or maybe lownull and highnull) and migrate to it instead of None in the long term.  I wonder how the PyData folks approach this. I don't remember a formal null-like value when I was swimming in that pond a year ago. There's http://pandas.pydata.org/pandas-docs/stable/missing_data.html, which in't very hepful.

jimfulton: Oh gawd, Python3 drops the cmp argument from sorted. whimper

jimfulton: I'm going to leave this the way it was.  Because a) Not sure what the right answer is, and I don't want to have to add special comparison at the C level.

jimfulton:
I believe I've addressed all of the excellent comments. Thanks!

jimfulton:
The lack of working CI is frustrating.

mauritsvanrees:
Awesome. Thanks a lot!

I see you made a release 4.4.0 with this, but I don't see a commit or tag for that. Do you need to push changes?

jimfulton:
Curses, caught again! Pushed. Thanks.

PR #62, opened 2017-01-16
Use macpython everywhere.

jamadden:
Add 3.6 for mac.

See zopefoundation/persistent#59

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
         - os: linux
           python: pypy
         - os: linux
-          python: pypy3
+          python: pypy3.3-5.2-alpha1
         # It's important to use 'macpython' builds to get the least
         # restrictive wheel tag. It's also important to avoid
         # 'homebrew 3' because it floats instead of being a specific version.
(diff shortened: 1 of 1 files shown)


mgedmin:
Very interesting Travis failures (Python 3.6.0 on Mac OS X):
```
    Download error on https://pypi.python.org/simple/persistent/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:749) -- Some packages may not be found!
```

I've restarted the job, to see if the error is transient.

jamadden:
I don't think it's transient, I think I may have already restarted that job once. It also [shows up for zope.interface](https://travis-ci.org/zopefoundation/zope.interface/jobs/192342208) 😢 

But I don't know what the problem is. I can't reproduce it locally with the macpython installed python, but I'm on OS X 10.12. I vaguelly recall some SSL changes in OS X Python 3.6, I'll see if I can find them.

mgedmin:
One possible avenue of inspection is to check whether the downloads fail when using pip (which uses requests, which I trust to do the right thing wrt TLS), or just when using setuptools (which uses stdlib's urllib).  Note that pip installing some packages may fall through to setuptools-downloads when those packages have a `setup_requires`.

Can you reproduce the SSL error locally on your Mac with Python 3.6?

jamadden:
Appveyor is green. Travis is green for one of the builds, six+ hours later, and almost green for the other (pending only one OS X build). The merge build of the previous zope.interface commit was still hung up on the OS X builds, and it was blocking other builds (notably Acquisition and ExtensionClass and Persistence) from proceeding. So I've cancelled that build and will merge this one, and if the merge build doesn't happen in a reasonable time I will probably cancel it too.

It seems reasonable to carefully consider how absolutely necessary automated OS X wheels are, given that the few projects that use them completely gum up the works for all the rest of the zopefoundation projects, and given how easy it is to have a compiler working on OS X (the UI will even prompt you the first time something invokes 'make' or 'cc', etc, IIRC).

PR #76, opened 2017-09-15
Fix possibility of rare crash during dealloc. Fixes #75

### LINKED ISSUE #75: Fix process of reference count during GC
When we subtract reference counter of an object, the destructor is automatically called if the value of counter is 0. 
At the same time an error occurs, as the GC executed while the destructor is still in running phase.
To eliminate the error, we have added below program which exclude objects from GC in the destructor in line 4.

static void
Wrapper_dealloc(Wrapper *self)
{
  PyObject_GC_UnTrack((PyObject *)self); //Add this line to eliminate self from GC
  Wrapper_clear(self);
  self->ob_type->tp_free((PyObject*)self);
}

Patch: [ZODB3-3.10.7.zip](https://github.com/zopefoundation/BTrees/files/1301526/ZODB3-3.10.7.zip)

jamadden:


jamadden:
Thanks! 

I know not all of Travis is green as I write this, but it takes *forever* for that to happen while we wait for the mac builds and docker builds to go. The main builds are all green both on Travis and Appveyor---and I tested on a mac, so I feel comfortable pushing the merge button.

PR #80, opened 2018-03-07
Support PURE_PYTHON at runtime. 

### LINKED ISSUE #78: BTrees should support the PURE_PYTHON environment variable at runtime
This has become common for zopefoundation packages such as [persistent](https://github.com/zopefoundation/persistent/blob/master/persistent/__init__.py#L21), [Acquisition](https://github.com/zopefoundation/Acquisition/blob/master/src/Acquisition/__init__.py#L945), [zope.security](https://github.com/zopefoundation/zope.security/blob/master/src/zope/security/proxy.py#L383) and so on. It's convenient for debugging and exploring the system, and also for testing.

BTrees does *not* support this variable, and this causes problems because it conflicts with `persistent`, which does:
```console
$ pip freeze | egrep  "BTree|pers"
BTrees==4.4.1
persistent==4.2.4.2
$ cat test.py
import persistent
from BTrees.OOBTree import OOBTree

class BaseClass(persistent.Persistent):
    pass

class MyTree(OOBTree, BaseClass):
    pass
$ python --version
Python 3.6.2
$ python test.py # C Extensions work
$ PURE_PYTHON=1 python test.py # Mixed does not
  File "test.py", line 7, in <module>
    class MyTree(OOBTree, BaseClass):
TypeError: multiple bases have instance lay-out conflict
```

If this is agreeable, I can work up a PR when I get some time.

jamadden:
Add failing tests and then fix them. Use a common function to DRY in
all the modules.

Fixes #78

Also use zope.testrunner in tox (and travis) because it allows passing
arguments to select what tests to run, see slow tests, etc.

Simplify setup.py to always build the C modules and avoid polluting
wheel caches (the same as zope.security and other projects that have
been worked on lately).

Bump the minimum 'persistent' requirement to one that actually works
in PURE_PYTHON mode (plus fix wheel metadata; otherwise we need
environment markers).

Test PURE_PYTHON in environments on Travis. Do some minor tuning to
eliminate redundant test cases and speed up test time.

### DIFF
--- a/BTrees/_compat.py
+++ b/BTrees/_compat.py
@@ -11,15 +11,20 @@
 # FOR A PARTICULAR PURPOSE
 #
 ##############################################################################
+import os
 import sys
+
+PYPY = hasattr(sys, 'pypy_version_info')
+# We can and do build the C extensions on PyPy, but
+# as of Persistent 4.2.5 the persistent C extension is not
+# built on PyPy, so importing our C extension will fail anyway.
+PURE_PYTHON = os.environ.get('PURE_PYTHON', PYPY)
+
 
 if sys.version_info[0] < 3: #pragma NO COVER Python2
 
     PY2 = True
     PY3 = False
-
-    from StringIO import StringIO
-    BytesIO = StringIO
 
     int_types = int, long
     xrange = xrange
@@ -38,16 +43,10 @@ if sys.version_info[0] < 3: #pragma NO COVER Python2
     def _ascii(x):
         return bytes(x)
 
-    def _u(s, encoding='unicode_escape'):
-        return unicode(s, encoding)
-
 else: #pragma NO COVER Python3
 
     PY2 = False
     PY3 = True
-
-    from io import StringIO
-    from io import BytesIO
 
     int_types = int,
     xrange = range
@@ -67,7 +66,34 @@ else: #pragma NO COVER Python3
     def _ascii(x):
         return bytes(x, 'ascii')
 
-    def _u(s, encoding=None):
-        if encoding is None:
-            return s
-        return str(s, encoding)
+
+def import_c_extension(mod_globals):
+    import importlib
+    c_module = None
+    module_name = mod_globals['__name__']
+    assert module_name.startswith('BTrees.')
+    module_name = module_name.split('.')[1]
+    if not PURE_PYTHON:
+        try:
+            c_module = importlib.import_module('BTrees._' + module_name)
+        except ImportError:
+            pass
+    if c_module is not None:
+        new_values = dict(c_module.__dict__)
+        new_values.pop("__name__", None)
+        new_values.pop('__file__', None)
+        new_values.pop('__doc__', None)
+        mod_globals.update(new_values)
+    else:
+        # No C extension,
+        # make the Py versions available without that extension
+        for py in [k for k in mod_globals.keys() if k.endswith('Py')]:
+            mod_globals[py[:-2]] = mod_globals[py]
+
+    # Assign the global aliases
+    prefix = module_name[:2]
+    for name in ('Bucket', 'Set', 'BTree', 'TreeSet'):
+        mod_globals[name] = mod_globals[prefix + name]
+
+    # Cleanup
+    del mod_globals['import_c_extension']
--- a/BTrees/_compat.py
+++ b/BTrees/_compat.py
@@ -85,9 +85,10 @@ def import_c_extension(mod_globals):
         new_values.pop('__doc__', None)
         mod_globals.update(new_values)
     else:
-        # No C extension,
-        # make the Py versions available without that extension
-        for py in [k for k in mod_globals.keys() if k.endswith('Py')]:
+        # No C extension, make the Py versions available without that
+        # extension. The list comprehension both filters and prevents
+        # concurrent modification errors.
+        for py in [k for k in mod_globals if k.endswith('Py')]:
             mod_globals[py[:-2]] = mod_globals[py]
 
     # Assign the global aliases
--- a/BTrees/tests/common.py
+++ b/BTrees/tests/common.py
@@ -2302,7 +2302,6 @@ class ConflictTestBase(object):
         super(ConflictTestBase, self).setUp()
         _skip_if_pure_py_and_py_test(self)
 
-
     def tearDown(self):
         import transaction
         transaction.abort()
--- a/BTrees/tests/common.py
+++ b/BTrees/tests/common.py
@@ -11,34 +11,48 @@
 # FOR A PARTICULAR PURPOSE
 #
 ##############################################################################
+from __future__ import division
 
+import unittest
 import platform
 from unittest import skip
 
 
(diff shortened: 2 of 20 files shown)


mgedmin:
> Simplify setup.py to always build the C modules and avoid polluting
wheel caches (the same as zope.security and other projects that have
been worked on lately).

But, presumably, if the C module build fails, `setup.py install` goes on ahead so you can still use BTrees on a machine without a C compiler?

jamadden:
> But, presumably, if the C module build fails, setup.py install goes on ahead so you can still use BTrees on a machine without a C compiler?

Yes. It's the same code that's used to do that in zope.interface, zope.security, zope.proxy, etc. I'm trying to get all our C extensions standardized to do that as well as respect PURE_PYTHON at runtime.

BTrees/tests/testBTrees.py
```diff
@@ -361,7 +358,7 @@ def test_LP294788(self):
         for id in ids:
             del t[id]
         #release all refs
-        ids = obj = id = None
+        ids = id = None
```
mgedmin: I'm assuming `obj` was otherwise unused in this method?

BTrees/tests/testBTreesUnicode.py
```diff
@@ -29,33 +29,31 @@ def setUp(self):
         #setup an OOBTree with some unicode strings
         from BTrees.OOBTree import OOBTree
         from BTrees._compat import _bytes
-        from BTrees._compat import _u
 
-        self.s = _u(b'dreit\xe4gigen', 'latin1')
+        self.s = b'dreit\xe4gigen'.decode('latin1')
```
mgedmin: This is equivalent to `u'dreit\xe4gigen'`.

I expect there was an explicit `latin1` because `_u('...\xe4...')` did not work.

BTrees/tests/testBTreesUnicode.py
```diff
@@ -29,33 +29,31 @@ def setUp(self):
...
 
         self.data = [(b'alien', 1),
                      (b'k\xf6nnten', 2),
                      (b'fox', 3),
                      (b'future', 4),
                      (b'quick', 5),
                      (b'zerst\xf6rt', 6),
-                     (_u(b'dreit\xe4gigen','latin1'), 7),
+                     (b'dreit\xe4gigen'.decode('latin1'), 7),
```
mgedmin: (Same here)

setup.py
```diff
@@ -15,17 +15,50 @@
...
+            self._unavailable(e)
+
+    def _unavailable(self, e):
+        print('*' * 80)
+        print("""WARNING:
+        An optional code optimization (C extension) could not be compiled.
+        Optimizations for this package will not be available!""")
+        print()
+        print(e)
+        print('*' * 80)
```
mgedmin: If you `print()` on Python 2, it'll print the repr of the empty tuple.

Please either use `print("")` or `from __future__ import print_function`.

mgedmin: One thing I remember about `pip` is that it hides the output of `setup.py` , so end-users will not know their cached wheels lack C modules...

It would be nice if we could make `python setup.py bdist_wheel` **fail** unless PURE_PYTHON is set, in which case
1. pip will print the failure message to the user
2. pip will not cache a C-module-less wheel in ~/.cache
3. pip will fall back to running `python setup.py install` directly

setup.py
```diff
@@ -87,58 +121,47 @@ def BTreeExtension(family):
...
+          "Framework :: ZODB",
+          "Topic :: Database",
+          "Topic :: Software Development :: Libraries :: Python Modules",
+          "Operating System :: Microsoft :: Windows",
+          "Operating System :: Unix",
+      ],
       author="Zope Foundation",
       author_email="zodb-dev@zope.org",
-      url="http://packages.python.org/BTrees",
+      url="https://github.com/zopefoundation/BTrees",
```
mgedmin: :+1:

tox.ini
```diff
@@ -42,29 +31,19 @@ commands =
 [testenv:w_zodb]
 basepython =
     python2.7
-commands =
-    python setup.py -q test -q
 deps =
-    zope.interface
-    persistent
-    transaction
+	{[testenv]deps}
```
mgedmin: Indentation seems a tad excessive.  Is this a literal tab?

BTrees/tests/common.py
```diff
@@ -77,6 +94,38 @@ def _closeRoot(self, root):
...
+
+        __traceback_info__ = self._getTargetClass(), persistent.Persistent
+        type('Subclass', (self._getTargetClass(), PersistentSubclass), {})
+
+    def testPurePython(self):
+        import importlib
+        kind = self._getTargetClass()
+        class_name = kind.__name__
+        module_name = kind.__module__
+        module = importlib.import_module(module_name)
```
mgedmin: Presumably if you already have a class from the module, then the module itself is already in `sys.modules[module_name]`.

BTrees/tests/common.py
```diff
@@ -77,6 +94,38 @@ def _closeRoot(self, root):
...
+
+        # If we're in the C extension mode, our target class
+        # module still doesn't have the _ in it, but we should be able to find
+        # a Py class that's different
+
+        self.assertNotIn('_', module_name)
+        self.assertIs(getattr(module, class_name), kind)
+
+        if not PURE_PYTHON and 'Py' not in type(self).__name__:
+            self.assertIsNot(getattr(module, class_name + 'Py'), kind)
```
mgedmin: Trying to understand this test hurts my brain, so instead I'll just trust you. ;)

BTrees/_compat.py
```diff
@@ -67,7 +66,34 @@ def compare(x, y):
...
+    if not PURE_PYTHON:
+        try:
+            c_module = importlib.import_module('BTrees._' + module_name)
+        except ImportError:
+            pass
+    if c_module is not None:
+        new_values = dict(c_module.__dict__)
+        new_values.pop("__name__", None)
+        new_values.pop('__file__', None)
+        new_values.pop('__doc__', None)
```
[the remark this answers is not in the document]
jamadden: They don't, no, but if they did it should probably have priority.

BTrees/tests/testBTrees.py
jamadden: That's correct.

tox.ini
jamadden: Whoops, I think it was.

BTrees/tests/testBTreesUnicode.py
jamadden: So it is!

BTrees/tests/common.py
```diff
@@ -2234,6 +2298,11 @@ class ConflictTestBase(object):
 
     storage = None
 
+    def setUp(self):
+        super(ConflictTestBase, self).setUp()
+        _skip_if_pure_py_and_py_test(self)
+
+
```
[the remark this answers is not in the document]
jamadden: 😄

PR #84, opened 2018-03-09
Drop Python 3.3 and add Python 3.7. 

### LINKED ISSUE #81: Drop Python 3.3 (add Python 3.7?)
Travis currently only has 3.7.0a4+, but that should be good enough to test with for this package, I am guessing.

jamadden:
Fixes #81.


jamadden:
Thanks!

icemac:
@jamadden Thank you for working on this issue. Are there still things which should be done before a new release? Could you please create it or give me (`icemac`) the necessary rights on PyPI?

jamadden:
@icemac I'm happy to share access rights! One thing to note when making a release at this point, though: the manylinux1 docker images don't have a 3.7 build, and of course 3.7's not listed in our .manylinux-install.sh, so we won't be uploading 3.7 binary wheels just yet (even though 3.7b3 is supposed to be the final ABI). (Actually, because of the SSL issues, it's not clear to me that manylinux1 will ever support python 3.7; manylinux2010 ought to though. That's just my guessing, though, I don't recall seeing anything about that on distutil-sig or the manylinux tracker.)

icemac:
@jamadden Thank you for the rights and sharing your insights. I think we could live without a 3.7 linux wheel for this release until there is a clear way how to create it.

PR #92, opened 2018-08-08
python 3.7 binary wheels.

fgregg:
This PR directs travis and appveyor to build binary wheels for python 3.7

### DIFF
--- a/.manylinux-install.sh
+++ b/.manylinux-install.sh
@@ -7,7 +7,8 @@ for PYBIN in /opt/python/*/bin; do
     if [[ "${PYBIN}" == *"cp27"* ]] || \
        [[ "${PYBIN}" == *"cp34"* ]] || \
        [[ "${PYBIN}" == *"cp35"* ]] || \
-       [[ "${PYBIN}" == *"cp36"* ]]; then
+       [[ "${PYBIN}" == *"cp36"* ]] || \
+       [[ "${PYBIN}" == *"cp37"* ]]; then
         "${PYBIN}/pip" install -e /io/
         "${PYBIN}/pip" wheel /io/ -w wheelhouse/
 	rm -rf /io/build /io/*.egg-info
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,10 @@ matrix:
           env: TERRYFY_PYTHON='macpython 3.5'
         - os: osx
           language: generic
-          env: TERRYFY_PYTHON='macpython 3.6.0'
+          env: TERRYFY_PYTHON='macpython 3.6.1'
+        - os: osx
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[build-system]
+requires = ["setuptools",
+            "wheel",
+            "persistent"]
```
jamadden: Adding a pyprojet.toml enables build isolation, meaning that packages installed in the virtualenv are ignored. [Prior to pip 18.0](https://github.com/pypa/pip/pull/5336), only binary wheels were allowed for build-system.requires. Taken together, this means that people installing BTrees from source (most likely those on platforms that don't already have binary wheels, e.g., BSD), must have the absolute latest pip.

I also wonder if it has any effect on those of us testing BTrees and persistent locally in the same virtualenv. Will we have to remember to pass `--no-build-isolation` to `pip install -e .`?

mgedmin: Which pip version introduced support for `pyproject.toml`?  Surely older pips ignore it altogether, so there's only a brief range of versions where this could cause problems?  (And wouldn't pip install'ing persistent beforehand create a wheel in pip's local cache that could then be reused?)

I think if people on weird platforms want to upgrade to the latest BTrees, they can also upgrade to the latest pip.

jamadden: > Which pip version introduced support for pyproject.toml?

[10.0b1](https://pip.pypa.io/en/stable/news/#b1-2018-03-31) from March 31 of this year...a lot more recently than I thought.

> Surely older pips ignore it altogether, so there's only a brief range of versions where this could cause problems?

Probably true, now that 18.0 is actually released.

> And wouldn't pip install'ing persistent beforehand create a wheel in pip's local cache that could then be reused?

Hopefully?

> I think if people on weird platforms want to upgrade to the latest BTrees, they can also upgrade to the latest pip.

I would think so toe. I only mention it because every time I use a "new" setuptools feature in gevent, I get bug reports from people still using old versions (> 2 years). So I would suggest it's something that should explicitly be called out in the change notes.

fgregg:
@mgedmin, travis-ci still does not natively support "3.7" so I don't think I should make the requested change.

jamadden:
@fgregg It does, if you use 'dist: xenial' and 'sudo: true'. (See https://github.com/travis-ci/travis-ci/issues/9069 or look at zope.proxy for an example.) Binary wheels built with a pre-release of 3.7 that old are pretty much useless, anyway, the ABI changed quite late in the release cycle.

fgregg:
Thanks for the tip @jamadden and sorry that you had to repeat what @mgedmin said.

PR #93, opened 2018-08-10
Avoid spurious "failures" on Appveyor during releases

mgedmin:
PowerShell interprets any stderr output from a child process as an
error.  This means warnings from pip about using
not-the-very-latest-version-how-dare-you and warnings from twine about
uploading the *.tar.gz which was probably already uploaded already and
might not be a bitwise identical copy because what even is
reproducibility cause the Appveyor builds to fail, even if the actual
binary wheels we want have been built and uploaded successfully.

*pauses to breathe*

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,6 +22,7 @@ install:
   - pip install -e .[test,ZODB]
 
 build_script:
+  - python -m pip install -U pip
   - pip install wheel
   - python -W ignore setup.py -q bdist_wheel
 
@@ -33,6 +34,6 @@ artifacts:
     name: wheel
 
 deploy_script:
-  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload dist/* }
+  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload dist/*.whl }
 
 deploy: on
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,6 +34,6 @@ artifacts:
     name: wheel
 
 deploy_script:
-  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload dist/*.whl }
+  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload --skip-existing dist/* }
 
 deploy: on


mgedmin:
> LGTM, if we assume that the releasing person already uploaded the .tar.gz.

I usually do releases with `zest.releaser` which uploads the .tar.gz.

Perhaps we could use `twine upload --skip-existing`?

> The rebuild of the 4.5.1 tag and wheels is my fault, I had to repush to get Travis to build mac and manylinux wheels

Is this why the 4.5.1 tag is not on the master branch's history?  Personally I'd prefer if you forced-push directly to master for such situations, because tags hanging off to a side are a pain with git (e.g. git fetch won't fetch them, and so git tag -l won't list them, and all my release automation that assumes every release has a tag breaks).

Is there any reasonable way you could prevent such accidents in the future?  E.g. local zest.releaser configuration overrides to remove the `[skip ci]` bit in the `setup.cfg` of every zopefoundation package would be fine by me.

mgedmin:
FWIW I'm experimenting with `twine upload --skip-existing` in https://github.com/zopefoundation/zodbpickle/pull/39.

jamadden:
> I usually do releases with zest.releaser which uploads the .tar.gz.

Me too.

> Perhaps we could use twine upload --skip-existing?

That might be a good idea.

> Is this why the 4.5.1 tag is not on the master branch's history? 

I'm not sure what you mean. I see the 4.5.1 tag in the drowdown list on github and can browse into it (https://github.com/zopefoundation/BTrees/tree/4.5.1). It shows up in a 'git tag -l' on a fresh clone.

> Personally I'd prefer if you forced-push directly to master for such situations, because tags hanging off to a side are a pain with git (e.g. git fetch won't fetch them -- unless you do an explicit git fetch --tags -- and so git tag -l won't list them, and all my release automation that assumes every release has a tag breaks).

I *thought* all I did was move the tag to a new commit that I'd reworded.

> Is there any reasonable way you could prevent such accidents in the future? E.g. local zest.releaser configuration overrides to remove the [skip ci] bit in the setup.cfg of every zopefoundation package would be fine by me.

I just removed the offending configuration from my ~/.pypirc, so it won't happen again from me. We could try putting `extra-message = ` in the `[zest.releaser]` section of each `setup.cfg` that builds and deploys wheels in case others have a similar configuration.

jamadden:
> I'm not sure what you mean

Oh, I think I understand. 5bcc3de6b66052d60eec218ae9c1ce8519e1f66a is still in master with the `[skip ci]` text and isn't pointed to by the tag.

mgedmin:
The --skip-existing solution works fine: https://ci.appveyor.com/project/mgedmin/zodbpickle/build/1.0.48 is all green!

PR #95, opened 2018-09-14
Fix tests against persistent 4.4.

### LINKED ISSUE #94: BTrees can't pip install with Python 3.7.0
<strike>I think it has something to do with the new (18.0) release of pip, but I'm still investigating.</strike>

https://travis-ci.org/zopefoundation/BTrees/jobs/415400906

jamadden:
Stop being quite so strict in order to account for the differences between the C and Python implementations. But do test that we don't break the new features when we didn't already have a better form.

Fixes #94

Refs https://github.com/zopefoundation/persistent/issues/92

### DIFF
--- a/BTrees/tests/common.py
+++ b/BTrees/tests/common.py
@@ -365,6 +365,7 @@ class MappingBase(Base):
         t = self._makeOne()
         for i in range(5):
             t[i] = i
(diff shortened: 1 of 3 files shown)


jamadden:
Thanks!

PR #108, opened 2019-07-29
Fix segfault on repr(OOBucket([... non-Latin-1 thingies ...]))

### LINKED ISSUE #106: SIGSEGV in bucket_repr() with items that have a non-Latin-1 repr() on Python 3
I was trying to convert a Python 2 Data.fs to Python 3 with zodbupdate.  More specifically, I was poking around in zodbbrowser looking at the converted database searching for problematic pickles when I discovered this segfault on repr() of a BTree bucket:

```
$ gdb --args env3/bin/python3.5 env3/bin/zodbbrowser Data.fs
(gdb) run
Thread 2 "python3.5" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff2cc1700 (LWP 14546)]
PyBytes_AsString (op=op@entry=0x0) at Objects/bytesobject.c:1123
1123	                                const char *errors,
(gdb) bt
#0  PyBytes_AsString (op=op@entry=0x0) at Objects/bytesobject.c:1123
#1  0x00007ffff55bf67c in bucket_repr (self=0x7fffac5db448) at BTrees/BucketTemplate.c:1836
#2  0x00005555556099b4 in PyObject_Repr (v=0x7fffac5db448) at Objects/object.c:482
#3  0x000055555568469d in call_function (oparg=<optimized out>, pp_stack=0x7ffff2cbc3c0) at Python/ceval.c:4743
#4  PyEval_EvalFrameEx (f=f@entry=0x7fffec967248, throwflag=throwflag@entry=0) at Python/ceval.c:3289
...
```

The source at this point is

```c
1831	    {
1832	        return NULL;
1833	    }
1834	#ifdef PY3K
1835	    rb = PyUnicode_AsLatin1String(r);
1836 →	    rv = PyOS_snprintf(repr, sizeof(repr),
1837	                       "%s(%s)", Py_TYPE(self)->tp_name,
1838	                       PyBytes_AsString(rb));
1839	    Py_DECREF(rb);
1840	#else
```

and `rb` is 

```
(gdb) p rb
$1 = (PyObject *) 0x0
```

because it looks like PyUnicode_AsLatin1String failed.  The unicode object itself is

```
(gdb) p *r
$3 = {ob_refcnt = 3, ob_type = 0x5555557f2340 <PyUnicode_Type>}

(gdb) p *(PyUnicodeObject*)r
$5 = {_base = {_base = {ob_base = {ob_refcnt = 3, ob_type = 0x5555557f2340 <PyUnicode_Type>}, length = 609, hash = -1, state = {
        interned = 0, kind = 2, compact = 1, ascii = 0, ready = 1}, wstr = 0x0}, utf8_length = 0, utf8 = 0x0, wstr_length = 0}, data = {
    any = 0x6100270028005b, latin1 = 0x6100270028005b <error: Cannot access memory at address 0x6100270028005b>, ucs2 = 0x6100270028005b, 
    ucs4 = 0x6100270028005b}}
```

I believe there's a missing error check for `rb` being NULL.

mgedmin:
Fixes #106.

The bug was in not checking the return value of PyUnicode_AsLatin1String() and passing a NULL to PyOS_snprintf().

The fix uses PyUnicode_FromFormat() and avoids all the hassle of manual encoding/sprintf/decoding and even calling repr().

I've also added a regression test and tweaked `tox.ini` to set PYTHONFAULTHANDLER=1 to make the segfault visible -- without that all you see is the test runner exiting unexpectedly with no additional output (and you have to interpret "exited with code -11" as _died from a SIGSEGV_ yourself).

### DIFF
--- a/BTrees/BucketTemplate.c
+++ b/BTrees/BucketTemplate.c
@@ -1814,51 +1814,40 @@ static PyObject *
 bucket_repr(Bucket *self)
 {
     PyObject *i, *r;
-#ifdef PY3K
-    PyObject *rb;
-#endif
+#ifndef PY3K
     char repr[10000];
     int rv;
+#endif
 
     i = bucket_items(self, NULL, NULL);
     if (!i)
     {
         return NULL;
     }
+#ifdef PY3K
+    r = PyUnicode_FromFormat("%s(%R)", Py_TYPE(self)->tp_name, i);
+    Py_DECREF(i);
+    return r;
+#else
     r = PyObject_Repr(i);
     Py_DECREF(i);
     if (!r)
     {
         return NULL;
     }
-#ifdef PY3K
-    rb = PyUnicode_AsLatin1String(r);
-    rv = PyOS_snprintf(repr, sizeof(repr),
-                       "%s(%s)", Py_TYPE(self)->tp_name,
-                       PyBytes_AsString(rb));
-    Py_DECREF(rb);
-#else
     rv = PyOS_snprintf(repr, sizeof(repr),
                        "%s(%s)", Py_TYPE(self)->tp_name,
                        PyBytes_AS_STRING(r));
-#endif
     if (rv > 0 && (size_t)rv < sizeof(repr))
     {
         Py_DECREF(r);
-#ifdef PY3K
-        return PyUnicode_DecodeLatin1(repr, strlen(repr), "surrogateescape");
-#else
         return PyBytes_FromStringAndSize(repr, strlen(repr));
-#endif
     }
     else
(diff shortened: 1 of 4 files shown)


mgedmin:
The large stack allocation is inside a `#ifndef PY3K`, so it should already be saved.

Splitting the entire definition would probably improve readability.  (OTOH I can't wait for Jan 2020 to roll around so I can submit a PR ripping out Python 2 compatibility and all the ifdefs.)

Good catch on the changelog note!

PR #112, opened 2019-10-25
Some documentation love

azmeuk:
Some little changes on the documentation.

- Internal :class:xxxx links are fixed;
- The main menu is clearer;
- Doc is python3 compatible;
- persistent links do link to the persistent documentation
- tox documentation use the readthedocs theme


docs/conf.py
```diff
@@ -250,4 +250,8 @@
 
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'http://docs.python.org/': None}
+intersphinx_mapping = {
+    "python": ('http://docs.python.org/', None),
+    "persistent": ('https://persistent.readthedocs.io/en/latest/', None),
+    "ZODB": ("http://www.zodb.org/en/latest/", None),
```
mgedmin: Can we use https:// everywhere?

tox.ini
```diff
@@ -36,9 +36,11 @@ deps =
 
 [testenv:docs]
 basepython =
-    python2.7
+    python3.7
 commands =
     sphinx-build -b html -d docs/_build/doctrees docs docs/_build/html
     sphinx-build -b doctest -d docs/_build/doctrees docs docs/_build/doctest
+deps =
+    --requirement doc-requirements.txt
```
mgedmin: Heh, I didn't even know that `-r` had a long form alternative.

docs/conf.py
azmeuk: Done for python.org, but there is a problem with https://www.zodb.org/

azmeuk: [Bug reported](https://github.com/zopefoundation/ZODB/issues/284)

mgedmin: Maybe we want to use https://zodb.readthedocs.io/en/latest/ instead?

azmeuk: Well, my understanding is that www.zodb.org is the official documentation, and zodb.readthedocs.io is a community book based on [this repository](https://github.com/cguardia/ZODB-Documentation), the latter received few love lately. The last commit is 4 years old, and most activity stop since 2011.

mgedmin: Oops, how confusing!  I wish it was called zodb-book.rtd.org or something then.

PR #114, opened 2019-11-06
Add support for Python 3.8

mgedmin:
I'm skipping the MacOS build on Travis because 3.8 is not available yet from https://github.com/MacPython/terryfy

Needed for #113.

PR #115, opened 2019-11-06
Restructure Travis to always build manylinux wheels

mgedmin:
This PR makes it so we _always_ build manylinux wheels, but only upload them for git tags.  This is so we can notice problems like https://github.com/zopefoundation/persistent/issues/117#issuecomment-550260284 _before_ making releases, instead of after.

The diff is messy, for which I apologize.  I had to reformat the YAML to understand it.  Highlights:

- Moved regular Linux Python builds outside build matrix, for conciseness (unfortunately this means 3.6 builds are now done twice: once without WITH_COVERAGE and once with)

- Move twine upload from .manylinux.sh into .travis.yml's after_success, next to the twine upload used for Mac OS builds

- This means Travis can no longer do an 'exit 0' in before_install, so I had to override the install and script sections too

- Move 'docker pull' into the install section so it's not so very empty

- Unify the two calls to twine upload to provide the twine password via environment variables instead of command-line or configuration files

- Put related commands inside a single multiline if statement instead of repeating the conditions

This branch is based on top of #114 so you only need review the last commit.

### DIFF
--- a/.manylinux.sh
+++ b/.manylinux.sh
@@ -2,8 +2,4 @@
 
 set -e -x
 
-docker pull $DOCKER_IMAGE
-
-docker run --rm -v `pwd`:/io $DOCKER_IMAGE $PRE_CMD /io/.manylinux-install.sh
-
-pip install twine && twine upload -u zope.wheelbuilder -p $PYPIPASSWORD wheelhouse/*
+docker run --rm -v "$(pwd)":/io $DOCKER_IMAGE $PRE_CMD /io/.manylinux-install.sh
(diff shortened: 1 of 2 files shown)


mgedmin:
i686 manylinux builds fail because cffi lacks a i686 manylinux 3.8 wheel on pypi: https://bitbucket.org/cffi/cffi/issues/433/i686-manylinux1-wheel-for-python-38-is

mgedmin:
There is now a cffi-1.13.2-cp38-cp38-manylinux2010_i686.whl on PyPI, let's see if that works...

mgedmin:
There is now a cffi-1.13.2-cp38-cp38-manylinux1_i686.whl on PyPI and it fixes our builds.

mgedmin:
I suppose I should rebase this now that #114 was merged.

PR #123, opened 2020-02-26
Remove most all the boilerplate required to add new trees

jamadden:
Follow on from #122. There were lots of .py files that have to be copied and then search-n-replaced.

This PR uses Python code to programmatically create the correct modules and test modules based on a declarative description of the data types. I'm undecided on leaving the `test_*.py` files around. On the one hand, they make zope-testrunner's `-m` filter work; on the other hand, they're more near-duplicate files.

There are four remaining custom test cases for integer overflow that I'd like to unify into the base cases.

I discovered and fixed a few bugs during the course of this, where test modules had drifted apart and didn't test everything; notably, Bucket and BTree behaved differently since 4.3.2 with regards to KeyError and TypeError in query methods.


jamadden:
I think this is ready for review. It's based on #122 so only the final commit is relevant.

jamadden:
There are a handful of cleanups that can be enabled by this and land in a subsequent PR. Most notably, I think the nasty way we're dealing with pickling to make sure we always pickle and unpickle the best implementation can probably be cleaned up.

icemac:
Rerun the failing TravisCI jobs as `persistent == 4.6.1` fixes them.

PR #126, opened 2020-03-16
Developper documentation improvements

azmeuk:
Moved development.txt in the docs directory so it is built alongside the rest of the documentation.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -313,8 +313,8 @@ invariant still holds, with ``lo==i`` and ``hi==lo+1==i+1``::
 so ``i`` is again the correct answer.
 
 
-Optimization points:
---------------------
+Optimization points
+-------------------
 
 + Division by 2 is done via shift rather via "/2".  These are signed ints, and
   almost all C compilers treat signed int division as truncating, and shifting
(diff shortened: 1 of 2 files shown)


docs/development.rst
```diff
@@ -1,20 +1,19 @@
-=====================
```
jamadden: I believe the adornments were correct to start with, as defined by the tools that ship with docutils.
```lisp
(defcustom rst-preferred-adornments '((?= over-and-under 1)
				      (?= simple 0)
				      (?- simple 0)
				      (?~ simple 0)
				      (?+ simple 0)
				      (?` simple 0)
				      (?# simple 0)
				      (?@ simple 0))
```

azmeuk: I just moved the titles levels one step down so it is coherent with the rest of the documentation.
Keeping this level would mess a bit with the other titles.

jamadden: No, I don't think it would. The rest of the documents use standard adornments. E.g., `overview.rst`:
```rest
==========
 Overview
==========
```

`api.rst`:
```rest
===============
 API Reference
===============


Protocol APIs
=============

.. automodule:: BTrees.Interfaces
```

jamadden: Cool, thanks! 

But it looks like now the commits are all confused 😭

azmeuk: Are you sure? I just rebased against master.

azmeuk: Ok that should be fixed now.

PR #128, opened 2020-03-18
docstrings for iterkeys, itervalues & iteritems

azmeuk:


jamadden:
As `keys`, `items`, and `values` all already return lazy sequences that do not materialize their data, and are the consistent standard available across `dict` and BTrees in all versions of Python, I'm not sure I see the need to document the `iter*` versions. I think those methods are there only for backwards compatibility and their use in new code should not be encouraged.

mgedmin:
IMHO they should be documented as such: deprecated methods, for backwards compatibility, use these other methods instead.

(It would probably be best to introduce a new interface, IKeyedDeprecated for them.  I would also like to move `has_key` there.)

jamadden:
zope.interface has two interfaces for those things, tied together in [IExtendedReadMapping](https://zopeinterface.readthedocs.io/en/latest/api/common.html#zope.interface.common.mapping.IExtendedReadMapping). Perhaps the objects here should just extend that interface.

(Perhaps the interfaces here could use a more thorough review to see where they fit with the new common collections interfaces in zope.interface 5.0. But `IExtendedReadMapping` is available before that.)

PR #130, opened 2020-03-18
Added internal documentation links

azmeuk:
So references are clickable when they are documented.

### DIFF
--- a/BTrees/check.py
+++ b/BTrees/check.py
@@ -16,7 +16,7 @@ Utilities for working with BTrees (TreeSets, Buckets, and Sets) at a low
 level.
 
 The primary function is check(btree), which performs value-based consistency
-checks of a kind btree._check() does not perform.  See the function docstring
+checks of a kind ``BTree._Tree._check()`` does not perform.  See the function docstring
 for details.
 
 display(btree) displays the internal structure of a BTree (TreeSet, etc) to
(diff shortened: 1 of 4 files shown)


BTrees/check.py
```diff
@@ -442,17 +442,17 @@ def visit_bucket(self, obj, path, parent, is_mapping,
 def check(btree):
     """Check internal value-based invariants in a BTree or TreeSet.
 
-    The btree._check() method checks internal C-level pointer consistency.
-    The check() function here checks value-based invariants:  whether the
+    The :meth:`~BTrees._base.btree._Tree._check` method checks internal C-level pointer consistency.
```
jamadden: `_base` and `_Tree` are implementation details that aren't published anywhere (and which only apply to the Python implementation). It would be fine to just express this as code:
```
The ``btree._check()``…
```

docs/overview.rst
```diff
@@ -180,11 +180,11 @@ exclusive of the range's endpoints.
 
 
 Each of the modules also defines some functions that operate on BTrees --
-:func:`~BTrees._base.difference`, :func:`~BTrees._base.union`, and :func:`~BTrees._base.intersection`.  The
-:func:`~BTrees._base.difference` function returns a Bucket, while the other two methods return
+:func:`~BTrees.Interfaces.IMerge.difference`, :func:`~BTrees.Interfaces.IMerge.union`, and :func:`~BTrees.Interfaces.IMerge.intersection`.  The
```
jamadden: 👍

azmeuk:
`tox -e docs` does not work for you?

jamadden:
> tox -e docs does not work for you?

Let me clarify: When I say "not able", I mean "I'm too lazy." Instead of being a branch in this repo, the PR is a merge from your own forked repo. I'd either have to clone it separately, or set up a different upstream and pull from that instead. (Yes, I know that [hub](https://hub.github.com) has shortcuts to make that easier, but, you know, lazy.)

azmeuk:
Ah ah:) No problem!

mgedmin:
@jamadden 

    git config --global alias.pr '!f() { a=${1:?usage: git pr NNN}; git fetch origin "+refs/pull/$a/head:refs/remotes/origin/pull/$a" && git checkout "pull/$a"; }; f'

and now you can do `git pr 130`.  You're welcome!

(just don't `git push` the fake pull/130 branch, it would create a real one)

azmeuk:
RTD has not been updated. Do you usually trigger builds manually? If so, there is a [RTD hook](https://docs.readthedocs.io/en/stable/webhooks.html#github) for github that could help!

mgedmin:
I have admin access to GitHub repo settings, but not RTD. Everything looks correct from the GitHub side: there's a webhook, it has all the right events enabled, and there are green checkmarks next to all recent deliveries:
Ekrano nuotrauka iš 2020-03-19 11-34-50

Someone who access to RTD will have to investigate this one.

jamadden:
RTD *has* been updated appropriately. The webhook caused this commit to build. [Here's the log](https://readthedocs.org/projects/btrees/builds/10636467/).

mgedmin:
Hm, I see RTD is still using Python 2.7?  Maybe it's about time to migrate to Python 3?

Also, I was mistaken: I do have RTD admin access to BTrees.

jamadden:
> I do have RTD admin access to BTrees.

I added you once I realized I was the single point of failure on it.

PR #133, opened 2020-03-20
[WIP] Backward browsing

azmeuk:
I would like to:

- implement a `reverse` parameter on `keys`, `values`, `items`, `iterkeys`, `itervalues`, `iteritems` methods.
- implement a reversed iterator so `reversed` would work on BTrees, Buckets, TreeSets and Sets.
- allow slices steps of -1 on `BTreeItems`.

Here is a first attempt to implement the `reverse` parameter on BTrees and Buckets. I prefer to check with you if I am going the right way before I spend too much time on this.

Basically, when `reverse` is True, we start the iteration with the last bucket, and then go backward. To do this I set a `prev` pointer on buckets, that work the same way than `next` but is not stored in the state.

### DIFF
--- a/BTrees/BTreeItemsTemplate.c
+++ b/BTrees/BTreeItemsTemplate.c
@@ -29,6 +29,9 @@
  * alone, or 'i'items (key+value pairs).  'kind' is also readonly after
  * initialization.
  *
+ * reverse determines wether to return values from start ot end or end to
+ * start.
+ *
  * The combination of currentbucket, currentoffset and pseudoindex acts as
  * a search finger.  Offset currentoffset in bucket currentbucket is at index
  * pseudoindex, where pseudoindex==0 corresponds to offset first in bucket
@@ -52,6 +55,7 @@ typedef struct
     int first;              /* Start offset in firstbucket    */
     int last;               /* End offset in lastbucket       */
     char kind;              /* 'k', 'v', 'i'                  */
+    int reverse;            /* Wether ot iterate backward     */
 } BTreeItems;
 
 #define ITEMS(O)((BTreeItems*)(O))
@@ -59,7 +63,8 @@ typedef struct
 static PyObject *
 newBTreeItems(char kind,
               Bucket *lowbucket, int lowoffset,
-              Bucket *highbucket, int highoffset);
+              Bucket *highbucket, int highoffset,
+              int reverse);
 
 static void
 BTreeItems_dealloc(BTreeItems *self)
@@ -145,6 +150,9 @@ BTreeItems_seek(BTreeItems *self, Py_ssize_t i)
     currentbucket = self->currentbucket;
     if (currentbucket == NULL)
         goto no_match;
+
+    if (self->reverse)
+        i = -i-1;
 
     delta = i - pseudoindex;
     while (delta > 0) /* move right */
@@ -335,6 +343,7 @@ BTreeItems_slice(BTreeItems *self, Py_ssize_t ilow, Py_ssize_t ihigh)
     Bucket *highbucket;
     int lowoffset;
     int highoffset;
+    int reverse = 0;
     Py_ssize_t length = -1;  /* len(self), but computed only if needed */
 
     /* Complications:
@@ -405,7 +414,7 @@ BTreeItems_slice(BTreeItems *self, Py_ssize_t ilow, Py_ssize_t ihigh)
         highoffset = self->currentoffset;
     }
     return newBTreeItems(self->kind,
-                         lowbucket, lowoffset, highbucket, highoffset);
+                         lowbucket, lowoffset, highbucket, highoffset, reverse);
 }
 
 static PyObject *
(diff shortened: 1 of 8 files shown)


jamadden:
Won't setting up `prev` pointers require traversing through all of the buckets to reach the end? And since this is setting an attribute, won't that require unghosting the objects? So this loads the entire tree structure into memory?

What's the advantage over just using `reversed()`?

azmeuk:
Indeed the idea is to not having to load the entire tree in memory, so `prev` will need to be stored in the state (it is just that, at the moment the test I wrote just test the behavior with a fresh BTree so the `prev` pointer is set each time there is a new bucket).
This is an interesting point to check with you however. Is it a problem to store an additional `prev` pointer in the state? I suppose BTrees without the `prev` pointer won't benefit from the *lazy* `reversed` behavior, at least in a first time (until the pointer is set everywhere).

freddrake:
There are some concerns.

1. Setting prev in the state of existing buckets can change a read-only transaction to a read-write transaction, which can be a substantial effect for an application.

2. The new state will not be compatible with older versions of BTrees, making it difficult to share a database between applications (yes, this happens), or even roll out the change gradually when many instances of an application share a database.

This would require a major-number bump at the least, along with significant visible warning in the documentation, change notes, and availability announcements.

Which doesn't make it a bad idea; there are certainly times this would have been helpful.

BTrees/BTreeItemsTemplate.c
```diff
@@ -29,6 +29,9 @@
  * alone, or 'i'items (key+value pairs).  'kind' is also readonly after
  * initialization.
  *
+ * reverse determines wether to return values from start ot end or end to
```
freddrake: Typo: "start ot end" --> "start to end"

jamadden:
>> 1. Setting prev in the state of existing buckets can change a read-only transaction to a read-write transaction, which can be a substantial effect for an application.

Our application forbids that entirely (read-only transactions are aborted and never committed in production; in testing, read-only transactions that attempt to make changes generally produce errors. [Unless explicitly allowed.]). So that's a non-starter for us.

>> 2. The new state will not be compatible with older versions of BTrees, making it difficult to share a database between applications (yes, this happens), or even roll out the change gradually when many instances of an application share a database.

Our environment requires both of those things (the second more than the first, but the first happens; it also happens when attempting to debug issues from different deployed versions). Stopping the world and causing user downtime to run what promises to be a very long migration is not something we like to have to do.

> Is it a problem to store an additional prev pointer in the state? 

When it comes down to it, I can't think of a situation where we really needed to lazily iterate a BTree in reverse. If the ability had been available, maybe we would have used it. But since it isn't, it's not something we think of doing. (The most common time we need to sort results is *after* they've been filtered and limited by other means.)

It's also not something that, to me at this time, seems worth paying the migration costs for.

azmeuk:
I am not totally sure a big migration would be mandatory: We can patch past minor versions so they do not crash when they meet a *new* state (with a `prev` pointer). I suppose this would be a harmless fix. It seems to me that ascending compatibility patches are pretty common technique is those situations. However that would still require to update the BTrees requirements on those application so they can benefit from those patches.

Then, this PR can either:
- have `reverse` features unavailable if the `prev` pointer is unavailable. That would raise an exception inviting the user to perform a migration to benefit this feature.
- have `reverse` features work in a degraded way (i.e. with slow performances) and raise python warnings to invite users to migrate.

Then we can make the migration mandatory for a future major release.

Reading BTrees in reverse is something my company often do, and we definitely would need better performances than `revesed(list(...))` currently offers. However I understand that you do not want to spent time on a feature you won't use, or you think can be dangerous.

How can we go further on this?
If I can provide a PR where no migration is needed, and a tested patch for past versions so they are not surprised to encounter a new BTrees state, would you consider to merge it? How does this kind of `design` decisions are usually taken in the zope community?

d-maurer:
Jason Madden wrote at 2020-3-23 04:56 -0700:
>>> 1. Setting prev in the state of existing buckets can change a read-only transaction to a read-write transaction, which can be a substantial effect for an application.
>
>Our application forbids that entirely (read-only transactions are aborted and never committed in production; in testing, read-only transactions that attempt to make changes generally produce errors. [Unless explicitly allowed.]). So that's a non-starter for us.

Something similar applies to Plone:
unexpected database writes are considered an indication for
a CSRF (= "Cross Side Request Forgery") attack.

d-maurer:
Jason Madden wrote at 2020-3-23 04:56 -0700:
> ...
>When it comes down to it, I can't think of a situation where we really needed to lazily iterate a BTree in reverse. If the ability had been available, maybe we would have used it. But since it isn't, it's not something we think of doing. (The most common time we need to sort results is *after* they've been filtered and limited by other means.)

I know of an application in `Products.AdvancedQuery`:
it supports lazy sorting of search results and could profit
from a lazy reverse listing of the keys of an index.
In `Products.ManagableIndex` I had supported this by
maintaining for an index a key list in reverse order.

However, I think that such a requirement is rare enough
that it could be provided at a different level
(as I did in `ManagableIndex`). It would help if
`BTrees` types could be made subclassable to facilitate the implementation
of additional features.

azmeuk:
Closing this PR for now, I will make a better patch when I have some time.

PR #134, opened 2020-03-20
Blacked the code

azmeuk:
I used [black](https://github.com/psf/black) to format the code.
I really like the code that blacks produces, and my stylecheckers too :), however I would understand if you think this is too much noise on the git history.

### DIFF
--- a/BTrees/Interfaces.py
+++ b/BTrees/Interfaces.py
@@ -16,7 +16,6 @@ from zope.interface import Interface, Attribute
 
 
 class ICollection(Interface):
-
     def clear():
         """Remove all of the items from the collection."""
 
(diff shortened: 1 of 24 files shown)


freddrake:
Perhaps I've missed it, but has there been a discussion about whether a tool like Black should be used on the zopefoundation repositories?  This feels like something that should be decided at a larger scale, especially given that there would be enormous churn.

The benefits should be well understood and the value agreed upon by active maintainers across the community.

@jamadden?  @mgedmin?

jugmac00:
There was a decision against the usage of black - for now.

https://github.com/zopefoundation/meta/issues/6

freddrake:
Closing this PR.

@azmeuk: Thank you for your interest in contributing!

azmeuk:
I totally understand!

PR #141, opened 2020-04-06
Fix #140 by turning OverflowError into TypeError.

### LINKED ISSUE #140: `key in btree` should ignore OverflowError
Currently, `key in btree` ignores TypeError, so this all works:
```pycon
>>> import BTrees
>>> bt = BTrees.family32.II.BTree()
>>> 'abc' in bt
False
>>> bt.get('abc') is None
True
```

But if the key is outside the range allowed, it fails:
```pycon
>>> x = 2 ** 64
>>> x in bt
OverflowError: Python int too large to convert to C long
>>>  bt.get(x)
OverflowError: Python int too large to convert to C long
```

For consistency, the `OverflowError` should be ignored in those cases too. (This would happen automatically if it was a `TypeError` instead of an `OverflowError`.) The pure-Python implementation already handles this (…by raising a TypeError, I think).

Also weird and should not raise (works in pure-Python):
```pycon
>>> bt.has_key(1)
0
>>> bt.has_key('abc')
KeyError: 'abc'
>>> bt.has_key(x)
OverflowError: Python int too large to convert to C long
```

Confirmed this to be an issue back to 4.6. [The changes in 4.3.2](https://btrees.readthedocs.io/en/latest/changes.html#id9) were not sufficient for the overflow case.

jamadden:
Fingers crossed for 32-bit builds on Windows.

I'd really prefer to make it raise a subclass `OutOfRangeError(TypeError,OverflowError)` with a detail, but that's annoyingly difficult in C...

### DIFF
--- a/BTrees/tests/common.py
+++ b/BTrees/tests/common.py
@@ -449,6 +449,19 @@ class MappingBase(Base):
         self.assertEqual(self._makeOne().get(1), None)
         self.assertEqual(self._makeOne().get(1, 'foo'), 'foo')
 
+    def testGetReturnsDefaultWrongTypes(self):
+        self.assertIsNone(self._makeOne().get('abc'))
+        self.assertEqual(self._makeOne().get('abc', 'def'), 'def')
+
+    def testGetReturnsDefaultOverflowRanges(self):
+        too_big = 2 ** 64 + 1
+        self.assertIsNone(self._makeOne().get(too_big))
+        self.assertEqual(self._makeOne().get(too_big, 'def'), 'def')
+
+        too_small = -too_big
+        self.assertIsNone(self._makeOne().get(too_small))
+        self.assertEqual(self._makeOne().get(too_small, 'def'), 'def')
+
     def testSetItemGetItemWorks(self):
         t = self._makeOne()
         t[1] = 1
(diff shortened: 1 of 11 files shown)


BTrees/tests/common.py
```diff
@@ -475,14 +488,24 @@ def testLen(self):
...
+    def testHasKeyOverflowAndTypes(self):
+        t = self._makeOne()
+
+        too_big = 2 ** 64 + 1
+        too_small = -too_big
+        self.assertNotIn(too_big, t)
+        self.assertNotIn(too_small, t)
+        self.assertFalse(t.has_key(too_big))
+        self.assertFalse(t.has_key(too_small))
+        self.assertFalse(t.has_key('abc'))
```
mgedmin, line 508: Should there be a `self.assertNotIn('abc', t')`?

BTrees/tests/testBTrees.py
```diff
@@ -114,14 +114,14 @@ def testBasicOps(self):
         t, keys = self._build_degenerate_tree()
         self.assertEqual(len(t), len(keys))
         self.assertEqual(list(t.keys()), keys)
-        # has_key actually returns the depth of a bucket.
```
mgedmin, line 117: Oh boy, I hope nobody relied on that! :)

mgedmin:
Apparently Python 3 on Windows still keeps raising OverflowError.

BTrees/tests/testBTrees.py
jamadden: The doc string says "Return a true value if the key is present, else a false value." while the example says "returns a true value, but exactly what undefined". So I hope not also :)

jamadden:
> Some of the exceptions feel a bit weird -- the "integer out of range" feels more like a ValueError than a TypeError to me. 

I agree, I don't particularly like the exceptions being raised. But the alternatives didn't seem much better than standardizing the exception the conversion functions raise (as `TypeError`).

1) Let the conversion function raise *either* `TypeError` or `OverflowError`. Modify every place that catches `TypeError` to also catch `OverflowError`. As the saying in computer science goes, 2 is a stupid number: that really should be a configurable list of exceptions per tree type, because those two don't really cover the gamut of options. (The hypothetical 8-byte (OID) tree or the 128-bit UUID tree might want to raise other exceptions.) And that's a lot of places to modify. If this were only Python and not C, that might be an option, but some sort of class or module constant for the exceptions is much more painful in C.
2) Catch and ignore *any* exception from the conversion functions. Easy in C, but [Just Say No](https://github.com/zopefoundation/zope.interface/pull/201).
3) Multiply-inherit from `TypeError` and `OverflowError` (and maybe `ValueError`). This is a variant of the solution here, standardizing on a single conversion error to catch, but allows for a more descriptive message and exception clauses (in user code). I like that, but it's hard to do in C; the easiest way I know is to import a Python module that declares the exception, and the use of inlined macros everywhere makes that a bit tricky. (Ultimately, I'd like to see what can be done using just a bit of C++ templates instead of macros; I suspect the results would be cleaner and easier to maintain and would make specializations like that a bit easier.)

jamadden:
Thanks!

PR #143, opened 2020-05-12
Don't allow to pass bad child into tree.__setstate__ to avoid memory corruption and crash

navytux:
Hello up there. To test my module that computes diff for BTrees I was playing
with manually creating BTrees with different topologies[1,2] and hit the
following bug that was leading to segmentation faults:

C implementation of Tree.__setstate__ allows to pass in arbitrary objects in
place of children and casts child to (Bucket*) if child type is not type of the
tree

    _without further checking that type of the child is actually Bucket_

This leads to crashes when later the code, that is accessing tree nodes,
goes to leafs, accesses passed in objects assuming they are buckets with
corresponding C-level Bucket structure layout, and oops dereferences e.g.
Bucket->keys, or Bucket->values from memory initialized via non-Bucket
C-level data.

-> Fix it by allowing to pass into tree.__setstate__ only children of
either tree or bucket types.

Note: for tree kind the type is checked exactly, because in many places C
implementation already does `if (SameType_Check(tree, X))` and assumes X is
of bucket kind if that check fails. For buckets we accept tree._bucket_type
subclasses as they are handled correctly and bucket type for tree.firstbucket
is already verified via "isinstance".

Kirill

P.S.

test___setstate___to_multiple_buckets is adjusted to avoid test failures
because Test_Tree._makeOne() was creating tree with ._bucket_type different
from _Bucket defined in that test.

[1] https://lab.nexedi.com/kirr/wendelin.core/blob/28010b7/wcfs/internal/xbtree.py
[2] https://lab.nexedi.com/kirr/wendelin.core/blob/28010b7/wcfs/internal/xbtree_test.py

### DIFF
--- a/BTrees/tests/common.py
+++ b/BTrees/tests/common.py
@@ -23,6 +23,7 @@ from unittest import skip
 from BTrees._compat import PY3
 from BTrees._compat import PURE_PYTHON
 from BTrees._compat import PYPY
+from BTrees._base   import _tp_name
 
 def _no_op(test_method):
     return test_method
@@ -1594,6 +1595,46 @@ class BTreeTests(MappingBase):
             s2 = pickle.dumps(loaded_one, proto)
             self.assertTrue(b'Py' not in s2)
             self.assertEqual(s2, s)
+
+    def testSetstateBadChild(self):
+        # tree used to allow to pass in non (tree or bucket) node as a child
+        # via __setstate__. This was leading to crashes when using C mode.
+        t = self._makeOne()
+        b = t._bucket_type()
+        if isaset(b):
+            b.add(1)
+        else:
+            b[1] = 11
+
+        # xchild is non-BTree class deriving from Persistent
+        import persistent
+        xchild = persistent.Persistent()
+        self.assertIs(xchild._p_oid, None)
+
+        typeErrOK = "tree child %s is neither %s nor %s" % \
+                        (_tp_name(type(xchild)), _tp_name(type(t)),
+                         _tp_name(t._bucket_type))
+
+        # if the following is allowed, e.g.
+        # t.__getstate__(), or t[0]=1 corrupt memory and crash.
+        with self.assertRaises(TypeError) as exc:
+            t.__setstate__(
+                    (
+                        (xchild,), # child0 is neither tree nor bucket
+                        b
+                    )
+                )
+        self.assertEqual(str(exc.exception), typeErrOK)
+
+        # if the following is allowed, e.g. t[5]=1 corrupts memory and crash.
+        with self.assertRaises(TypeError) as exc:
+            t.__setstate__(
+                    (
+                        (b, 4, xchild),
+                        b
+                    )
+                )
+        self.assertEqual(str(exc.exception), typeErrOK)
 
 
 class NormalSetTests(Base):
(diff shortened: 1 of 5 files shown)


BTrees/tests/common.py
```diff
@@ -1595,6 +1596,54 @@ def test_legacy_py_pickle(self):
...
+                    (
+                        (xchild,), # child0 is neither tree nor bucket
+                        b
+                    )
+                )
+        except TypeError as v:
+            self.assertEqual(str(v), typeErrOK)
+        else:
+            #t[0]=1
+            raise AssertionError("Expected exception")
```
jamadden, line 1631:
```suggestion
with self.assertRaises(TypeError) as exc:
    t.__setstate__(
            (
                (xchild,), # child0 is neither tree nor bucket
                b
            )
        )
self.assertEqual(str(exc.exception), typeErrOK)
```

I think that will take care of the coverage reduction.

BTrees/tests/common.py
```diff
@@ -1595,6 +1596,54 @@ def test_legacy_py_pickle(self):
...
+
+        # if the following is allowed, e.g. t[5]=1 corrupts memory and crash.
+        try:
```
jamadden, line 1634: `with self.assertRaises` as above.

navytux:
@jamadden thanks for feedback. I've amended the patch with the following interdiff

```diff
--- a/BTrees/tests/common.py
+++ b/BTrees/tests/common.py
@@ -1617,32 +1617,24 @@ def testSetstateBadChild(self):
 
         # if the following is allowed, e.g.
         # t.__getstate__(), or t[0]=1 corrupt memory and crash.
-        try:
+        with self.assertRaises(TypeError) as exc:
             t.__setstate__(
                     (
                         (xchild,), # child0 is neither tree nor bucket
                         b
                     )
                 )
-        except TypeError as v:
-            self.assertEqual(str(v), typeErrOK)
-        else:
-            #t[0]=1
-            raise AssertionError("Expected exception")
+        self.assertEqual(str(exc.exception), typeErrOK)
 
         # if the following is allowed, e.g. t[5]=1 corrupts memory and crash.
-        try:
+        with self.assertRaises(TypeError) as exc:
             t.__setstate__(
                     (
                         (b, 4, xchild),
                         b
                     )
                 )
-        except TypeError as v:
-            self.assertEqual(str(v), typeErrOK)
-        else:
-            #t[5]=1
-            raise AssertionError("Expected exception")
+        self.assertEqual(str(exc.exception), typeErrOK)
 
 
 class NormalSetTests(Base):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,9 @@
 4.7.3 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Fix ``Tree.__setstate__`` to no longer accept children besides
+  tree or bucket types to prevent crashes. See `PR 143
+  <https://github.com/zopefoundation/BTrees/pull/143>`_ for details.
 
 
 4.7.2 (2020-04-07)
```

navytux:
Thanks, @jamadden. What is the path forward for my patch to be merged?

jugmac00:
@navytux - in order to contribute to any Zope projects, you have to sign a contribution agreement.

Then you get commit rights and you can and should merge yourself.

Here is the link:
https://www.zope.org/developer/becoming-a-committer.html

Please follow the Plone link within mentioned page for further information.

navytux:
@jugmac00, thanks for clarifying. I contacted management. Hopefully I could come back with all neccessary papers, but let's see.

navytux:
I have signed contributor agreement and sent it to agreements@plone.org.

navytux:
@jamadden, @jugmac00, @polyester, thanks; I've applied my patch to master.

PR #144, opened 2020-05-29
Shortcut operators for union, intersection and difference

### LINKED ISSUE #142: Implement bitwise operators
The `union`, `intersection` and `difference` functions have not their bitwise operators equivalent implemented:
```python
>>> from BTrees.OOBTree import OOTreeSet, union
>>> union(OOTreeSet({1, 2}), OOTreeSet({2, 3}))
OOSet([1, 2, 3])

>>> OOTreeSet({1, 2}) & OOTreeSet({2, 3})
TypeError: unsupported operand type(s) for &: 'BTrees.OOBTree.OOTreeSet' and 'BTrees.OOBTree.OOTreeSet'

unsupported operand type(s) for &: 'BTrees.OOBTree.OOTreeSet' and 'BTrees.OOBTree.OOTreeSet'
```

I suggest implementing `__or__`, `__and__` and `__xor__` methods, and make them call `union`, `intersection` and `difference`. I volunteer to provide a patch.

What do you think?

azmeuk:
Fixes #142 
Implementation of `__sub__`,  `__or__` and `__and__` as shortcuts for `difference`, `union` and `intersection`.

This makes the usage of BTrees datastructures closer to native python sets, that support those operators.

### DIFF
--- a/BTrees/Interfaces.py
+++ b/BTrees/Interfaces.py
@@ -320,16 +320,6 @@ class IMerge(Interface):
         collections.
         """
 
-    def __and__(self, other):
-        """Shortcut for :meth:`~BTrees.Interfaces.IMerge.intersection`"""
-
-    def __or__(self, other):
-        """Shortcut for :meth:`~BTrees.Interfaces.IMerge.union`"""
-
-    def __sub__(self, other):
-        """Shortcut for :meth:`~BTrees.Interfaces.IMerge.difference`"""
-
-
 
 class IBTreeModule(Interface):
     """These are available in all modules (IOBTree, OIBTree, OOBTree, IIBTree,
--- a/BTrees/Interfaces.py
+++ b/BTrees/Interfaces.py
@@ -126,24 +126,24 @@ class IKeySequence(IKeyed, ISized):
 
 
 class ISet(IKeySequence, ISetMutable):
-    def __and__(self, other):
+    def __and__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.intersection`"""
 
-    def __or__(self, other):
+    def __or__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.union`"""
 
-    def __sub__(self, other):
+    def __sub__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.difference"""
 
 
 class ITreeSet(ISetMutable):
-    def __and__(self, other):
+    def __and__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.intersection`"""
 
-    def __or__(self, other):
+    def __or__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.union`"""
 
-    def __sub__(self, other):
+    def __sub__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.difference"""
 
 
@@ -282,13 +282,13 @@ class IBTree(IDictionaryIsh):
               key=generate_key()
         """
 
-    def __and__(self, other):
+    def __and__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.intersection`"""
 
-    def __or__(self, other):
+    def __or__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.union`"""
 
-    def __sub__(self, other):
+    def __sub__(other):
         """Shortcut for :meth:`~BTrees.Interfaces.IMerge.difference`"""
 
 
--- a/BTrees/Interfaces.py
+++ b/BTrees/Interfaces.py
@@ -277,6 +277,10 @@ class IMerge(Interface):
     :meth:`BTrees.IIBTree.IIBTree.union` can only be used with :class:`~BTrees.IIBTree.IIBTree`,
     :class:`~BTrees.IIBTree.IIBucket`, :class:`~BTrees.IIBTree.IISet`, and :class:`~BTrees.IIBTree.IITreeSet`.
 
+    The number protocols methods ``__and__``, ``__or__`` and ``__sub__`` are provided
+    by all the data structures. They are shortcuts for :meth:`~BTrees.Interfaces.IMerge.intersection`,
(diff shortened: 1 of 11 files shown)


BTrees/tests/common.py
```diff
@@ -2313,6 +2313,8 @@ def testUnion(self):
                 C = self.union(A, B)
                 self.assertTrue(not hasattr(C, "values"))
                 self.assertEqual(list(C), self._union(A, B))
+                if PY3 or A.__class__ == B.__class__:
+                    self.assertEqual(list(C), list(A | B))
```
jamadden, line 2317:
```suggestion
                    self.assertEqual(set(A) | set(B), set(A | B))
```

This seems like a more fair, direct comparison. Using `C` is relying on the results of `self.union`, which, I think, is not really what we're trying to test for here (and likewise in the rest of the methods); the fact that it's used internally is an implementation detail.

Also, perhaps that loosens the restrictions about needing PY3 and/or equal classes? In any case, if something is throwing a `TypeError`, it would be nice to have a test case showing that.

azmeuk: Your assertion suggestion sounds good indeed.

Actually I am not really sure what is going on with python2 here. Using those operators between different kinds of datastructures, say a Set and a TreeSet, sometimes result in a `TypeError: unsupported operand type(s)`, and sometimes not.

azmeuk: I could finally tighten the tests.
It seems that in python2, using those operators on two different C classes objects raises a TypeError. Any other configuration seems to work.

jamadden: Were you able to debug to the specific C call that’s failing? Is it before or after our code takes control?

Maybe we just need appropriate `__r*__` definitions?

I *think* my preference for the operators would be to be fairly loose in what they accept, especially if it smooths over any 2/3 differences.  Maybe that entails calling the class’s constructor on the argument when needed (for incompatible types, that probably results in a more specific exception)? Those who wish a very strict, fast interpretation can continue to use the existing functions. 

Of course it’s always easier to loosen restrictions rather than add them in the future. My main motivation for these new operators is to try to get the 2/3 semantics as much the same as possible. With `None` we initially went the strict Py3 way everywhere then reverted to the looser Py2 semantics. Here it seems good to go with the looser Py3 version everywhere from the start.

azmeuk: I think python2 type checks are related to [this flag](https://docs.python.org/2.7/c-api/typeobj.html#Py_TPFLAGS_CHECKTYPES).

jamadden: Nice discovery!

I'd like to suggest that in the future you try to avoid force-pushing to PRs that are being reviewed. Reviewers get a notification with the same commit message that they've gotten before. When they go to github to see the changes, the github review tools are crippled: they have to review the entire changeset again and try to remember what might be different this time vs the last time.

Moreover, I would argue that the history here, together with good commit messages, would have been extremely valuable. History would have been able to answer the question "Why is Py_TPFLAGS_CHECKTYPES set? What's the purpose?". Now, it's just there and we have to guess what part of the new tests failed without it.

azmeuk: Ok I see. Actually I have never really been this side of the patch so I am not familiar with the tools github provides. I will avoid rebase in those situations in the future.

BTrees/Interfaces.py
```diff
@@ -277,6 +277,10 @@ class IMerge(Interface):
     :meth:`BTrees.IIBTree.IIBTree.union` can only be used with :class:`~BTrees.IIBTree.IIBTree`,
     :class:`~BTrees.IIBTree.IIBucket`, :class:`~BTrees.IIBTree.IISet`, and :class:`~BTrees.IIBTree.IITreeSet`.
 
+    The number protocols methods ``__and__``, ``__or__`` and ``__sub__`` are provided
+    by all the data structures. They are shortcuts for :meth:`~BTrees.Interfaces.IMerge.intersection`,
+    :meth:`~BTrees.Interfaces.IMerge.union` and :meth:`~BTrees.Interfaces.IMerge.difference`.
+
```
azmeuk, line 283: I just set this line of documentation because I was not sure on which interface to add the new operators. Is it ok like this? Or should I add the operators to the IBTree, ITreeSet and ISet interfaces?

mgedmin:
All the OS X builders failed because

```
Checking dist/BTrees-4.7.3.dev0-cp38-cp38-macosx_10_9_x86_64.whl: FAILED
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 42: Error: Unknown interpreted text role "meth".
  warning: `long_description_content_type` missing.  defaulting to `text/x-rst`.
```

no Sphinx markup is allowed in README.rst that's used for long_description.

(In the past I've sometimes [resorted to a judicious str.replace()](https://github.com/mgedmin/objgraph/blob/90be2be02badc8f7b748e9cc1b27cca27678057c/setup.py#L17-L21) to overcome this problem.)

BTrees/Interfaces.py
jamadden: It's the modules (e.g., `BTrees.OOBTree`) that implement `IMerge`, so I don't think the method definitions of the operators belong here. It doesn't hurt anything to mention in the main docstring here that there are operator synonyms for the functions defined by this interface, but I think that the operator method definitions should be moved to the appropriate datatype interfaces.

CHANGES.rst
```diff
@@ -5,7 +5,11 @@
 4.7.3 (unreleased)
 ==================
 
-- Nothing changed yet.
+- BTrees, TreeSet, Set and Buckets implements the ``__and__``,
```
jamadden, line 11: 👍

jamadden:
Since the `long_description` error is just coming from `CHANGES.rst` directly, I personally would probably just remove the `:meth:` tags and just turn the references into double-backtick code.

BTrees/Interfaces.py
azmeuk: So, just to be sure, IBTree, ITreeSet and ISet ?

jamadden: That sounds right to me. (I don't think there's an `IBucket`.)

azmeuk:
Everything should be good now :)

PR #147, opened 2020-11-06
fix persistency bug in the Python version

### LINKED ISSUE #118: BTree "the bucket being iterated changed size"
Versions:
Python 3.6
```
BTrees==4.6.0
ZODB==5.5.1
```

I have a `BTrees.OOBTree.BTree` in a ZODB which seems to have somehow become corrupted.

While our system was running we observed that the data the BTree was returning seemed suspicious (while iterating the tree, items which should have been there were not actually there). We took a snapshot of the tree and manually examined it, and determined that the structure of the BTree seems to have become corrupted. 

When we try to iterate the tree (or its keys):
```
for key in tree.keys():
    print('key: {}'.format(key))
```
We get this error:
```
  File "/home/test.py", line 40, in <module>
    for key in tree.keys():
RuntimeError: the bucket being iterated changed size
```

Interestingly, as far as I can tell our live service was not getting this exception (but it definitely was also not returning all of the data it should have been) - I'm not sure why the discrepancy between what our live service was seeing with this DB versus what I see when I download it and look through it locally.

`BTrees.check.check` shows no issues.

Obviously, I am not modifying the tree while iterating over it in my debugging environment - but when I look deeper into the structure of the tree by iterating the buckets one by one I see this:
```
Bucket: BTrees.OOBTree.OOBucket([('0231ed5f-6a2a-4675-a75a-103c6818de95', <Data object at 0x7faba5d396d8 oid 0x265b9 in <Connection at 7fabaa263cc0>>), ('03697f69-500b-4f7f-8ff1-b05c7d85b6cc', <Data object at 0x7faba5d39748 oid 0x26480 in <Connection at 7fabaa263cc0>>), ('03ac8297-38bf-4148-bd4e-df43c1e3d08b', <Data object at 0x7faba5d397b8 oid 0x262b5 in <Connection at 7fabaa263cc0>>), ('03d44486-31a2-4ed7-be24-d755f3cbff2a', <Data object at 0x7faba5d39828 oid 0x25fdb in <Connection at 7fabaa263cc0>>)])
... (a bunch more buckets that all look fine)
Bucket: BTrees.OOBTree.OOBucket([('4480e440-2d21-45f8-a909-664a85330df7', <Data object at 0x7faba5d41978 oid 0x264ad in <Connection at 7fabaa263cc0>>), ('452fc9ad-2544-4385-a034-3120a69409e8', <Data object at 0x7faba5d419e8 oid 0x264c4 in <Connection at 7fabaa263cc0>>), ('4554ce23-1ef5-4ff3-bee2-eb9233243448', <Data object at 0x7faba5d41a58 oid 0x2612f in <Connection at 7fabaa263cc0>>), ('45f92417-c80f-481a-b0d7-60110d757591', <Data object at 0x7faba5d41ac8 oid 0x25ccf in <Connection at 7fabaa263cc0>>)])
Bucket: BTrees.OOBTree.OOBucket([])
Bucket: BTrees.OOBTree.OOBucket([])
Bucket: BTrees.OOBTree.OOBucket([])
Bucket: BTrees.OOBTree.OOBucket([])
Bucket: BTrees.OOBTree.OOBucket([('91660989-9343-4f72-a076-94c8fc6c155f', <Data object at 0x7faba5ce03c8 oid 0xf769 in <Connection at 7fabaa263cc0>>), ('9176559f-e650-47bf-812b-dc4e75eaf299', <Data object at 0x7faba5ce0438 oid 0x223fa in <Connection at 7fabaa263cc0>>), ('91cd51fc-e9bd-4730-843f-b2553a06e63d', <Data object at 0x7faba5ce04a8 oid 0x22e68 in <Connection at 7fabaa263cc0>>), ('91d1b461-77ba-42b6-930b-326d7b2badbf', <Data object at 0x7faba5ce0518 oid 0x26310 in <Connection at 7fabaa263cc0>>), ('92acce7e-4e15-45c0-b954-2c25f9325731', <Data object at 0x7faba5ce0588 oid 0x221b9 in <Connection at 7fabaa263cc0>>), ('92e5b508-84a1-404b-9692-da2a7696c762', <Data object at 0x7faba5ce05f8 oid 0x22d32 in <Connection at 7fabaa263cc0>>), ... )])
... (more buckets with data here)
```

From what I can tell, 0 size buckets in the "middle" of the bucket list is not expected and is what triggers the exception I'm getting. Additionally if you look closely, it seems like a large swath of buckets is lost, because there should be a relatively even distribution of GUIDs as keys (and I see that from first digit 0-4 and first digit 9-f), but first digit 5-8 is missing.

**My questions are:**
1. Why doesn't `BTrees.check.check` notice this?
2. What could cause this type of corruption?

It's possible that this is related to #68, although they look sufficiently different to me that I figured I would post a new issue. It is also possible that this only started happening in `BTrees==4.6.0` as we moved to that version relatively recently (we were on `BTrees==4.5.1` previously) - but given how rare this is occurring we may just not have noticed it until now...

d-maurer:
fixes #118

### DIFF
--- a/BTrees/tests/common.py
+++ b/BTrees/tests/common.py
@@ -91,24 +91,14 @@ class SignedMixin(object):
     KEY_RANDRANGE_ARGS = (-2000, 2001)
 
 
-class Base(SignedMixin):
-    # Tests common to all types: sets, buckets, and BTrees
+class ZODBAccess(object):
 
     db = None
-
-    def _getTargetClass(self):
-        raise NotImplementedError("subclass should return the target type")
-
-    def _makeOne(self):
-        return self._getTargetClass()()
-
-    def setUp(self):
-        super(Base, self).setUp()
-        _skip_if_pure_py_and_py_test(self)
 
     def tearDown(self):
         if self.db is not None:
             self.db.close()
+            del self.db
 
     def _getRoot(self):
         from ZODB import DB
@@ -130,6 +120,21 @@ class Base(SignedMixin):
         # "Cannot close connection joined to transaction"
         transaction.abort()
         root._p_jar.close()
+
+    
+class Base(ZODBAccess, SignedMixin):
+    # Tests common to all types: sets, buckets, and BTrees
+
+    def _getTargetClass(self):
+        raise NotImplementedError("subclass should return the target type")
+
+    def _makeOne(self):
+        return self._getTargetClass()()
+
+    def setUp(self):
+        super(Base, self).setUp()
+        _skip_if_pure_py_and_py_test(self)
+
 
     def testPersistentSubclass(self):
         # Can we subclass this and Persistent?
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@
 ==================
 
 - Fix persistency bug in the Python version
-  (`#118 https://github.com/zopefoundation/BTrees/issues/118`_).
+  (`#118 <https://github.com/zopefoundation/BTrees/issues/118>`_).
 
 - Fix ``Tree.__setstate__`` to no longer accept children besides
   tree or bucket types to prevent crashes. See `PR 143
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@
 4.7.3 (unreleased)
 ==================
 
+- Fix persistency bug in the Python version
+  (`#118 https://github.com/zopefoundation/BTrees/issues/118`_).
+
 - Fix ``Tree.__setstate__`` to no longer accept children besides
   tree or bucket types to prevent crashes. See `PR 143
   <https://github.com/zopefoundation/BTrees/pull/143>`_ for details.
--- a/tox.ini
+++ b/tox.ini
@@ -29,6 +29,12 @@ basepython =
 deps =
     ZODB
 
+[testenv:w_zodb-pure]
+basepython =
+    python2.7
+deps =
+    ZODB
+
 [testenv:coverage]
 basepython =
     python3.6
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist =
 # Jython support pending 2.7 support, due 2012-07-15 or so.  See:
 # http://fwierzbicki.blogspot.com/2012/03/adconion-to-fund-jython-27.html
 #   py27,jython,pypy,coverage,docs
-    py27,py27-pure,py35,py35-pure,py36,py37,py38,pypy,pypy3,w_zodb,coverage,docs
+    py27,py27-pure,py35,py35-pure,py36,py37,py38,pypy,pypy3,w_zodb,w_zodb-pure,coverage,docs
 
 [testenv]
 usedevelop = true
--- a/BTrees/_base.py
+++ b/BTrees/_base.py
@@ -1050,6 +1050,7 @@ class _Tree(_Base):
                     self._firstbucket = child._next
                     removed_first_bucket = True
             del data[index]
+            self._p_changed = True
 
         return removed_first_bucket, value
 
--- a/BTrees/tests/testPersistency.py
+++ b/BTrees/tests/testPersistency.py
@@ -1,0 +1,47 @@
+##############################################################################
+#
+# Copyright (c) 2020 Zope Foundation and Contributors.
+# All Rights Reserved.
+#
+# This software is subject to the provisions of the Zope Public License,
+# Version 2.1 (ZPL).  A copy of the ZPL should accompany this distribution.
+# THIS SOFTWARE IS PROVIDED "AS IS" AND ANY AND ALL EXPRESS OR IMPLIED
+# WARRANTIES ARE DISCLAIMED, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF TITLE, MERCHANTABILITY, AGAINST INFRINGEMENT, AND FITNESS
+# FOR A PARTICULAR PURPOSE
+#
+##############################################################################
+
+from unittest import TestCase
+
+from ..OOBTree import OOBTree, OOBTreePy
+from .common import _skip_wo_ZODB, ZODBAccess
+
+
+BUCKET_SIZE = OOBTreePy.max_leaf_size
+
+
+class TestPersistency(ZODBAccess, TestCase):
+    @_skip_wo_ZODB
+    def test_empty_bucket_persistency(self):
+        from transaction import commit
+        root = self._getRoot()
+        try:
+            # tree with 3 buckets (internal implementation details)
+            tree = OOBTree(
+                dict((i, i) for i in range(3 * BUCKET_SIZE // 2 + 2)))
+            root["tree"] = tree
+            commit()
+            # almost clear the second bucket keeping the last element
+            for i in range(BUCKET_SIZE // 2, BUCKET_SIZE - 1):
+                del tree[i]
+            commit()
+            del tree[BUCKET_SIZE - 1]  # remove the last element
+            commit()
+            tree._check()
+            tree._p_deactivate()
+            tree._check()  # fails in case of bad persistency
+        finally:
+            self._closeRoot(root)
+
+            


mgedmin (approved):
LGTM.

Do you think it would be hard to write a regression test for this case?

CHANGES.rst
```diff
@@ -5,6 +5,9 @@
 4.7.3 (unreleased)
 ==================
 
+- Fix persistency bug in the Python version
+  (`#118 https://github.com/zopefoundation/BTrees/issues/118`_).
```
mgedmin, line 9: This is invalid ReST syntax, which gets caught by our Mac OS jobs in Travis CI running `twine check`:

```
   `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 39: Error: Unknown target name: "#118 https://github.com/zopefoundation/btrees/issues/118".
```

You're missing the `<` `>` around the URL.

mgedmin (changes requested):

d-maurer:
Marius Gedminas wrote at 2020-11-6 16:02 -0800:
>@mgedmin requested changes on this pull request.
> ...
>You're missing the `<` `>` around the URL.

Thank you for this catch! Fixed.

d-maurer:
Marius Gedminas wrote at 2020-11-6 16:00 -0800:
>@mgedmin approved this pull request.
> ...
>Do you think it would be hard to write a regression test for this case?

The difficult part comes from the fact that only the Python implementation
is affected. Otherwise, a test would be simple:
construct a tree with 2 buckets; empty the second bucket;
commit (i.e. write to ZODB); reload; check.

Even when the C implementation is available, I can access
the Python implementation (via class names with `Py` suffix).
However, some magic seems to cause a class name change
which might affect storing and loading the tree
(I suppose that at least the C implementation classes are used
on reloading).

I can deactivate the C implmentation via the envvar `PURE_PYTHON`
(that's what I did in my manual test). However, the envvar
affects only the initial `BTrees` import. Therefore, this approach
is difficult in part of a test suite (where `BTrees` might
already have been imported for preceding tests).

I have seen in `tox.ini` that there are `*-pure` envs
which apparently set the envvar `PURE_PYTHON`.
But, the test above would require both `w_zodb` as well as
`-pure` and I have not seen a hint that they are available together.

mgedmin:
I don't think it would be good to target the pure-python implementation specifically; it'd be better to have a generic test for this case that gets run against both C and Python implementations.

We don't need to have a single test suite run cover all the implementations.  We have multiple CI jobs for a reason, and some of them run with PURE_PYTHON=1.

If we're missing a job with PURE_PYTHON and w_zodb in Travis CI, let's add one!  I don't particularly care about tox.ini, unless it would help writing that test to have an additional toxenv, in which case go right ahead!

d-maurer:
> If we're missing a job with PURE_PYTHON and w_zodb in Travis CI, let's add one! I don't particularly care about tox.ini, unless it would help writing that test to have an additional toxenv, in which case go right ahead!

I have added a test.

When I run the test in isolation (`PURE_PYTHON=1 python -m unittest discover -s BTrees -t . -p 'testPersist*.py'`) it correctly detects the problem (when the fix is not available) **BUT** `tox -e w_zodb-pure` does not. I suppose `tox` is not working as I expect: i.e. that a suffix `-pure` is not sufficient to get the envvar `PURE_PYTHON` set. Would be good if a `tox` expert could look into this.

jugmac00:
I have no time to lookup the exact syntax, but my gut feeling tells me this should work:
```
[testenv:{py35,w_zodb}-pure]
setenv =
    PURE_PYTHON=1
```

If you want to look it up yourself, search for "tox factors".

d-maurer:
Jürgen Gmach wrote at 2020-11-9 01:57 -0800:
>I have no time to lookup the exact syntax, but my gut feeling tells me this should work:
>```
>[testenv:{py35,w_zodb}-pure]
>setenv =
>    PURE_PYTHON=1
>```
>
>If you want to look it up yourself, search for "tox factors".

The current `tox.ini` contains:
```
[testenv]
...
setenv =
  ...
  pure: PURE_PYTHON=1
```
As I understand the documentation this should work but apparently does not.


I am currently searching a way to convince `tox` to show the
environ variables it is using to executes the commands.
Even a large number of `-v` options does not do this.

d-maurer:
The problem was not a missing `PURE_PYTHON` envvar but a missing dependency.

I assumed that the `tox` env `w_zodb-pure` would inherit the settings of env `w_zodb`. This is not the case. As a consequence `w_zodb-pure` lacked the `ZODB` dependency and the test (which requires the ZODB) was skipped.

I made several trials to cover both `w_zodb` and `w_zodb-pure` in a single section -- all failing. Finally, I used a copy of the `w_zodb` section to define `w_zodb-pure`.

BTrees/tests/common.py
```diff
@@ -91,24 +91,14 @@ class SignedMixin(object):
...
-        return self._getTargetClass()()
-
-    def setUp(self):
-        super(Base, self).setUp()
-        _skip_if_pure_py_and_py_test(self)
-
     def tearDown(self):
         if self.db is not None:
             self.db.close()
+            del self.db
```
mgedmin, line 101: Why this rather than, say, `self.db = None`?

mgedmin (approved):
lgtm

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist =
 # Jython support pending 2.7 support, due 2012-07-15 or so.  See:
 # http://fwierzbicki.blogspot.com/2012/03/adconion-to-fund-jython-27.html
 #   py27,jython,pypy,coverage,docs
-    py27,py27-pure,py35,py35-pure,py36,py37,py38,pypy,pypy3,w_zodb,coverage,docs
+    py27,py27-pure,py35,py35-pure,py36,py37,py38,pypy,pypy3,w_zodb,w_zodb-pure,coverage,docs
```
mgedmin, line 6: Do we have a .travis.yml job corresponding to this?

mgedmin: Yes, AFAICT _all_ of our Travis jobs include ZODB.

d-maurer:
Marius Gedminas wrote at 2020-11-9 07:41 -0800:
>@mgedmin approved this pull request.
>
>lgtm
>
>>      def tearDown(self):
>         if self.db is not None:
>             self.db.close()
>+            del self.db
>
>Why this rather than, say, `self.db = None`?

I wanted to restore the initial state (where `db` is defined
at class level). Of course `self.db = None` would have almost the same
effect.

d-maurer:
Marius Gedminas wrote at 2020-11-9 07:42 -0800:
>@mgedmin commented on this pull request.
>> @@ -3,7 +3,7 @@ envlist =
> # Jython support pending 2.7 support, due 2012-07-15 or so.  See:
> # http://fwierzbicki.blogspot.com/2012/03/adconion-to-fund-jython-27.html
> #   py27,jython,pypy,coverage,docs
>-    py27,py27-pure,py35,py35-pure,py36,py37,py38,pypy,pypy3,w_zodb,coverage,docs
>+    py27,py27-pure,py35,py35-pure,py36,py37,py38,pypy,pypy3,w_zodb,w_zodb-pure,coverage,docs
>
>Do we have a .travis.yml job corresponding to this?

I have still very few knowledge about travis
(and made my first steps with `tox`).

mgedmin:
> I wanted to restore the initial state (where `db` is defined
> at class level).

Ah!  I didn't realize that was the case

> Of course `self.db = None` would have almost the same effect.

and would maybe avoid a possible bug in the future if somebody moves the class attribute into `__init__` and doesn't realize they need to also update this code.

d-maurer:
Marius Gedminas wrote at 2020-11-9 13:51 -0800:
> ...
>and would maybe avoid a possible bug in the future if somebody moves the class attribute into `__init__` and doesn't realize they need to also update this code.

This is a `unittest.TestCase` class. Its `__init__` is (mostly) private
(and called for each individual test). Thus, there is no danger.

In fact, I added the `del` (only) to help me interactively debug why my test
(initially) did not reveal the problem. It is not important for
the real test suite (you can delete it if you like).

As to the problem: the persistency loophole is very small:
it affects only the case that a transaction deletes the last
item in a bucket without having previously deleted anything in this
bucket. My initial test had emptied the whole bucket in the same
transaction and this did not reproduce the problem.

mgedmin:
> This is a `unittest.TestCase` class. 

I should stop reviewing code when I'm tired.  I didn't even notice that!

I retract my comments on this.

Issue #151, opened 2021-03-12
Add support to release aarch64 wheels


odidev:
Problem
---------
On aarch64, pip install BTrees builds the wheels from source code and then install it. It requires user to have development environment installed on his system. also, it take some time to build the wheels than downloading and extracting the wheels from pypi.

Resolution
-----------
On aarch64, pip install BTrees should download the wheels from pypi

@d-maurer , please let me know your interest on releasing aarch64 wheels. I can help in this.

jamadden:
Fixed by #154

odidev:
Can you please let me know the plans for releasing the aarch64 wheels? Thanks.

PR #152, opened 2021-03-12
Add linux aarch64 wheel support

odidev:
Added linux aarch63 wheel support.
Related to #151. @d-maurer Could you please review this PR.

### DIFF
--- a/.manylinux-install.sh
+++ b/.manylinux-install.sh
@@ -2,6 +2,9 @@
 
 set -e -x
 
+# Missing libffi on aarch64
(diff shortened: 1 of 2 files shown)


icemac:
Thank you for your PR. Unfortunately we are no longer able to use TravisCI (See https://github.com/zopefoundation/meta/issues/38). We are currently on the way to switch to GitHub Actions. When this is done all packages which include C code should provide aarch64 wheels.

Meanwhile I am closing this PR.

PR #154, opened 2021-03-26
Initial port to GitHub Actions

jamadden:
As used in persistent and zope.interface.

Unlike those, enable coverage conditionally, it's far too slow to always run (builds go from a minute to 6 minutes).

Adds Python 3.9 and aarch64 wheels.

PR #157, opened 2021-03-30
Move BTrees to src/

### LINKED ISSUE #155: Can we move to the src/ layout?
As with https://github.com/zopefoundation/persistent/issues/158

I was just bitten by one of the problems with having the directory in the root while trying to bring up the manylinux aarch64 builds. (Long story short, 'pip install BTrees' and using a git clone as the working directory are incompatible — the local directory shadows the installed version. You wind up only getting the pure-Python implementation because the shared libraries can't be found in `./BTrees/`.)

jamadden:
Fixes #155

I neglected to add 3.9 to appveyor.yml in #154 so I also did that here.


icemac (approved):
LGTM from reading the code.
This was way more work than I would have expected. Thank you for tackling this.

jamadden:
Thanks!

PR #158, opened 2021-03-31
Let PURE_PYTHON=0 force the use of C extensions.

### LINKED ISSUE #156: Support PURE_PYTHON=0 to require C extensions
Like `zope.interface` does. 

This is handy for testing on CI to be sure we're getting the extensions and to fail early if we're not. 

Currently, if we expect the extensions, but don't actually have them, we fail late in the test run with some confusing errors about `<class ABC> is <class ABC>`.

jamadden:
Except on PyPy.

Fixes #156

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,3 +7,4 @@ relative_files = True
 exclude_lines =
     pragma: no cover
     class I[A-Z]\w+\((Interface|I[A-Z].*)\):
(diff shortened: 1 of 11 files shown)


icemac (approved):
LGTM from reading over the diff.

jamadden:
Thank you!

PR #159, opened 2021-03-31
Make objects actually implement interfaces and update interface definitions

### LINKED ISSUE #138: Interface definitions outdated
For example, they refer to `__nonzero__` and `__getslice__` on all versions of Python.

I suspect many of the low-level interfaces defined here could be replaced by the corresponding interfaces from `zope.interface.common.collections`.

cf #128

jamadden:
- Make the BTree objects (BTree, TreeSet, Set, Bucket)  of each module actually provide the interfaces defined in  BTrees.Interfaces. Previously, they provided no interfaces.

- Update the definitions of `ISized` and `IReadSequence` to simply  be `zope.interface.common.collections.ISized` and  `zope.interface.common.sequence.IMinimalSequence` respectively.

- Remove the `__nonzero__` interface method from ICollection. No  objects actually implemented such a method; instead, the boolean value  is typically taken from `__len__`.

Fixes #138 

(Currently based on #156 to avoid a conflict; only the last commit is relevant here.)


src/BTrees/Interfaces.py
```diff
@@ -13,35 +13,28 @@
...
-        """
-
-    def __getslice__(index1, index2):
-        """Return a subsequence from the original sequence.
 
-        The subsequence includes the items from index1 up to, but not
-        including, index2.
-        """
+# Backwards compatibility alias. To be removed in 5.0
+IReadSequence = IMinimalSequence
```
icemac, line 37: Should the scheduled removal use a deprecation warning to inform users to update their code?

src/BTrees/Interfaces.py
```diff
@@ -52,9 +45,9 @@ def has_key(key):
         """
 
     def keys(min=None, max=None, excludemin=False, excludemax=False):
-        """Return an :class:`~BTrees.Interfaces.IReadSequence` containing the keys in the collection.
+        """Return an :class:`~BTrees.Interfaces.IMinimalSequence` containing the keys in the collection.
```
icemac, line 48: Maybe this should point to `zope.interface.common.sequence.IMinimalSequence` as the actual interface. (Same for the other changes previously referencing `IReadSequence`.)

src/BTrees/tests/common.py
```diff
@@ -128,13 +128,36 @@ class Base(ZODBAccess, SignedMixin):
...
+        verifyObject(self._getTargetInterface(), t)
+
+        if hasattr(t, 'keys'):
+            from BTrees.Interfaces import IReadSequence
+            keys = t.keys()
+            if type(keys) not in (tuple, list):
+                verifyObject(IReadSequence, keys)
+
+        if hasattr(t, 'values'):
+            from BTrees.Interfaces import IReadSequence
```
icemac, line 157: Hm, I thought `IReadSequence` was the deprecated name, shouldn't this here (and above) be `IMinimalSequence`?

icemac (approved):
I like the idea, but there are some open questions.

src/BTrees/Interfaces.py
jamadden: Great question!

In theory, yes, it probably should. That would require adding a dependency on `zope.deprecation`, which this package currently does not have. This package has the extremely minimal requirements of just `persistent` and `zope.interface`, and it seems desirable to me to keep the requirements minimal as this is such a low-level package.

So, I'm hoping we can get away with just docs-deprecation. From a practical/pragmatic standpoint, since nothing actually implemented any of these interfaces, there would be little incentive to use them. Searching GitHub finds essentially no references to "BTrees.Interfaces" or "IReadSequence" outside of copies of this package (there's a handful of references no the the `BTreesConflictError` defined here, but that seems to be it).

What do you think?

src/BTrees/tests/common.py
jamadden: Good point.

src/BTrees/Interfaces.py
jamadden: OK.

src/BTrees/Interfaces.py
icemac: Thank you for the research on the usage of these interfaces. I am okay with deprecating only in the documentation.

jamadden:
Thank you!

PR #160, opened 2021-04-05
Make multiunion(), union(), difference(), and intersection() work with arbitrary iterables

### LINKED ISSUE #24: multiunion error on C but pass on python
I met a problem about multiunion function that when using default C implementation it reports error but could work correctly when using python implementation... For example

```
fwd_index = OOBTree()
tp = (1,2,3)
fwd_index["a"] = tp

IF.multiunion(self._fwd_index.values(("a",))
```

jamadden:
The Python implementation already did this, so this only updates the C version to match.

For multiunion, all arguments can be arbitrary iterables. The others derive their return type from the first argument, so it must still be a BTree object, but the second can be anything. (And technically, this limitation only applies to the python version; the C version actually accepts anything for both arguments, except for `difference` which also has the same limitation. I wasn't motivated to fix the Python version to accept arbitrary items in the first argument. But I will open a bug report if this gets merged.)

Fixes #24

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,14 @@
 4.8.0 (unreleased)
 ==================
 
+- Make the ``multiunion``, ``union``, ``intersection``, and
+  ``difference`` functions accept arbitrary Python iterables (that
+  iterate across the correct types). Previously, the Python
+  implementation allowed this, but the C implementation only allowed
+  objects (like ``TreeSet`` or ``Bucket``) defined in the same module
+  providing the function. See `issue 24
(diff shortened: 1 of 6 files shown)


jamadden:
> I wasn't motivated to fix the Python version to accept arbitrary items in the first argument.

I found the motivation; the second commit fixes the python implementation. It turned out that the infrastructure was all already there, it just wasn't being used.

d-maurer (approved):
A feature I dearly missed

jamadden:
Thank you!

PR #163, opened 2021-04-07
Python 2: Forbid type objects as keys.

### LINKED ISSUE #153: Python 2: OxBTrees allow types as keys; Python 3 does not
Consider:
```python
class Old:
    pass

class New(object):
    pass

from BTrees import family64
bt = family64.OO.BTree()
```

On Python 3, both `Old` and `New`, which are types, cannot go in the tree because they have default comparison:
```pycon
>>> bt[Old] = 1
>>> bt[New] = 1
```

Under Python 2, however, the default comparison is only detected for old-style classes; new style `type` apparently doesn't have default comparison that we can detect:
```pycon
>>> bt[Old] = 1
>>> bt[New] = 1
```

Type ordering in Python 2 [is based on address](https://github.com/python/cpython/blob/8d21aa21f2cbc6d50aab3f420bb23be1d081dac4/Objects/typeobject.c#L669-L679), making it not suitable for use in a BTree. It emits a warning if you run Python with the `-3` argument.

Can we try harder (or just special case) to find that `type` objects shouldn't be used as keys?

jamadden:
Fixes #153.

Also take the opportunity to clean up the TODOs in _datatypes.py with regards to checking for default comparison: Use a metaclass so we get caching, and only check for the special methods on the type object.

### DIFF
--- a/src/BTrees/_datatypes.py
+++ b/src/BTrees/_datatypes.py
@@ -157,9 +157,17 @@ class Any(DataType):
 
 class _HasDefaultComparison(ABC):
     """
-    An ABC for checking whether an item has default comparison.
+    An `ABC <https://docs.python.org/3/library/abc.html>_` for
+    checking whether an item has default comparison.
 
-    This automatically gets us some caching.
+    All we have to do is override ``__subclasshook__`` to implement an
+    algorithm determining whether a class has default comparison.
+    Python and the ABC machinery will take care of translating
+    ``isinstance(thing, _HasDefaultComparison)`` into something like
+    ``_HasDefaultComparison.__subclasshook__(type(thing))``. The ABC
+    handles caching the answer (based on exact classes, no MRO), and
+    getting the type from ``thing`` (including mostly dealing with
+    old-style) classes on Python 2.
     """
 
     # Comparisons only use special methods defined on the
--- a/src/BTrees/_datatypes.py
+++ b/src/BTrees/_datatypes.py
@@ -20,6 +20,13 @@ from operator import index as object_to_index
 from struct import Struct
 from struct import error as struct_error
 
+try:
+    from abc import ABC
+except ImportError:
+    # Python < 3.4
+    from abc import ABCMeta
+    ABC = ABCMeta('ABC', (object,), {})
+
 from ._compat import int_types
 from .utils import Lazy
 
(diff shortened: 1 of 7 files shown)


src/BTrees/_datatypes.py
```diff
@@ -20,6 +20,13 @@
 from struct import Struct
 from struct import error as struct_error
 
+try:
+    from abc import ABC
+except ImportError:
+    # Python < 3.4
```
icemac, line 26: Maybe mention Python 2.7 here as it is the only version < 3.4 which is currently supported.
This could ease dropping Python 2 support later on when it is appropriate.

icemac (commented):
I have to remove myself from the reviewers list as I do not understand what's going on here.

src/BTrees/_datatypes.py
jamadden: No problem, thanks for the suggestion.

jamadden:
> I have to remove myself from the reviewers list as I do not understand what's going on here.

I apologize for being unclear. I admit to mixing in the fix for #153 ("Python 3 forbids classes as keys, but Python 2 doesn't") with some semi-unrelated cleanup.

The direct fix for #153 was to specifically check `type` and forbid it as a key, or, if it's a subclass of `type` (a metaclass) forbid it if the metaclass doesn't implement comparison. This happened by default on Python 3, but I think it's valuable to add for those migrating from Python 2 — there's nothing worse than BTree corruption!

It turns out there are a lot of differences among the supported implementations of Python as to how "doesn't implement comparison" can be detected. The refactoring changes both simplified and improved the speed of that. (I hope.)

Please let me know if I can help clarify further. I'd very much like to get this merged and released.

src/BTrees/_datatypes.py
```diff
@@ -161,58 +242,9 @@ def as_value_type(self):
...
-        if (lt is None
-                # TODO: Shouldn't we only check __cmp__ on Python 2?
-                # Python 3 won't use it.
-                and getattr(item, '__cmp__', None) is None):
-            raise TypeError("Object has default comparison")
-
-
     def __call__(self, item):
-        self._check_default_comparison(item)
+        if isinstance(item, _HasDefaultComparison):
```
icemac, line 246: `_HasDefaultComparison` is defined as a new class above but where is `item` registered for it, so this check can become true?

jamadden: > _HasDefaultComparison is defined as a new class above

Right. Specifically, it's defined as a new abstract base class, i.e., something that extends `abc.ABCMeta`. (We could do this with a regular `type` subclass, but extending `ABCMeta` gets us caching of the results of `isinstance` automatically.)

> where is `item` registered for it, so this check can become true?

That's the beauty part (as the saying goes). By making `_HasDefaultComparison` implement `__subclasshook__`, we [algorithmically define](https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__) what `isinstance(item, _HasDefaultComparison)` means. No need to register anything! The algorithm is the same as what was here before, just moved into `__subclasshook__`. (The expression `isinstance(item, Cls)` basically boils down to `Cls.__subclasscheck__(type(item))`; `ABCMeta` defines `__subclasscheck__` to do some caching, and if the answer is not in the cache, to call `Cls.__subclasshook__()`.)

jamadden: Did that help clear things up sufficiently?

icemac: TIL, I think I now understood how it works. Maybe adding a link to the Python documentation would help future readers of the code who rarely use abstract base classes.

icemac (approved):

src/BTrees/_datatypes.py
jamadden: Done!

jamadden:
Thank you!

PR #164, opened 2021-04-08
Register BTree/Bucket as MutableMapping and Set/TreeSet as MutableSet, add missing methods/operators

### LINKED ISSUE #121: BTrees should register as `collections.abc.MutableMapping` and `MutableSet`
The ABCs for `Mapping` and `MutableMapping` don't define a `__subclasshook__`, so even though BTrees have all the correct methods, `isinstance(btree, Mapping)` is never true.

It would be nice if the BTree module itself called `MutableMapping.register(…)` so that client code doesn't have to do that.

The same goes for `MutableSet`. It probably also applies for `KeysView`, `ItemsView` and `ValuesView` (though I've never found a need for those myself).

jamadden:
Add tests to verify they implement the required methods. Turns out they didn't, so implement them.
This means `popitem` for the mappings, and `pop/discard/isdisjoint` plus a swath of operators for the sets.

Do this in both C and Python. The implementations are probably on the naive side, in that they work with generic iterators and don't try to do any complex structural merging, but that makes them easier to understand and hopefully be correct. Performance shouldn't be too bad, nothing worse than `O(N*logM)` (roughly). The `__iand__` method requires extra storage space of up to `O(N)`, and `__xor__` requires storage of O(N*M).

Fixes #121

This was a lot bigger than I expected.

### DIFF
--- a/src/BTrees/_base.py
+++ b/src/BTrees/_base.py
@@ -617,7 +617,9 @@ class Bucket(_MutableMappingMixin, _BucketBase):
 class _MutableSetMixin(object):
     # Like _MutableMappingMixin, but for sets.
     def isdisjoint(self, other):
-        'Return True if two sets have a null intersection.'
+        """
+        Return True if two sets have a null intersection.
+        """
         for value in other:
             if value in self:
                 return False
--- a/src/BTrees/_base.py
+++ b/src/BTrees/_base.py
@@ -90,7 +90,30 @@ class _Base(Persistent):
 
         _BTree_reduce_up_bound = _BTree_reduce_as
 
-class _BucketBase(_Base):
+class _ArithmeticMixin(object):
+    def __sub__(self, other):
+        return difference(self.__class__, self, other)
+
+    def __rsub__(self, other):
+        return difference(self._set_type, type(self)(other), self)
+
+    def __or__(self, other):
+        return union(self._set_type, self, other)
+
+    __ror__ = __or__
+
+    def __and__(self, other):
+        return intersection(self._set_type, self, other)
+
+    __rand__ = __and__
+
+    def __xor__(self, other):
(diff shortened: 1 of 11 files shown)


src/BTrees/_base.py
```diff
@@ -582,7 +614,68 @@ def merge_output(it):
     def __repr__(self):
         return self._repr_helper(self.items())
 
-class Set(_BucketBase):
+class _MutableSetMixin(object):
+    # Like _MutableMappingMixin, but for sets.
+    def isdisjoint(self, other):
+        'Return True if two sets have a null intersection.'
```
icemac, line 620:
```suggestion
        """Return True if two sets have a null intersection."""
```

icemac (approved):
LGTM for the Python code. I cannot judge the C code.

src/BTrees/_base.py
jamadden: Amusingly enough, that’s a copy from the Python standard library! They’re inconsistent too.

jamadden:
> LGTM for the Python code. I cannot judge the C code.

I firmly admit to not being a C API expert. I'm at best a journeyman, and I am responsible for [at least one true memory leak](https://github.com/zopefoundation/zope.interface/issues/216) in zope.interface. That leak involved garbage collection, not regular methods.

Here, I've done my best to keep the C implementations simple, and they don't involve type-level GC changes, so I hope that's a point in favor of this PR :) Also, these are mostly new methods and operators, so even if there are some memory leaks, they won't be sudden and widespread — hopefully people can help report them.

I *think* I can mostly rule out obvious memory misuse errors, as they would result in segfaults; I know, I experienced them several times during development. 

As always, more eyeballs are welcome, so I will wait at least several hours before merging (it's after my bedtime in my timezone anyway). 

Thank you!

PR #167, opened 2021-05-19
Expose max_internal_size and max_leaf_size as rw on the C classes

### LINKED ISSUE #166: Python/C Inconsistency: Getting/Setting max_internal_size on the BTree class
For awhile, BTrees have [supported the ability to customize their node sizes](https://btrees.readthedocs.io/en/latest/overview.html#btree-node-sizes). 

The only way to know what the default values are is to look at that documentation, though, because the C implementation uses a static C module-level integer to store them. 

```pycon
>>> from BTrees.OOBTree import OOBTree, OOBTreePy
>>> OOBTree.max_internal_size
AttributeError: type object 'BTrees.OOBTree.OOBTree' has no attribute 'max_internal_size'
```

The only way to use different values is to define and use an entirely new class, as in the example:
```pycon
>>> import BTrees.OOBTree
>>> class MyBTree(BTrees.OOBTree.BTree):
...     max_leaf_size = 500
...     max_internal_size = 1000
```

This is because the C classes don't allow setting any attributes:

```pycon
>>> OOBTree.max_internal_size = 42
TypeError: can't set attributes of built-in/extension type 'BTrees.OOBTree.OOBTree'
```

This can be a particular problem when you want to customize the node sizes across an application, using code you don't control (for example, `zope.index`, `zc.catalog`, etc, all directly create BTree objects — there is a workaround for this using the `family` argument, but it's pretty hacky). It's also tricky of you want to change the sizes used by existing persistent tree objects, because you can't change their class.

The Python implementation doesn't suffer from either of these problems:
```pycon
>>> OOBTreePy.max_internal_size
250
>>> OOBTreePy.max_leaf_size
30
>>> OOBTreePy.max_internal_size = 42
```

The first problem — reading the values — is easy to solve. We just need to copy them into the class dict at module init time.

That leaves writing the values. The `type` object is what's throwing the error (for good reason). This can be solved by using a metaclass for the C BTree class. If we're careful about what attributes we allow to be set in our `tp_setattro`, we shouldn't have any problems. I have a prototype implementation that seems to work (all tests pass).

Another plus would be that we would bypass the sidetable lookup that `zope.interface` does for builtin classes by allowing it to store `__implemented__` in the type dictionary.

If there's interest, I can bring the prototype up to quality. The biggest downside I see is the use of a metaclass because of its confusing nature. I don't think the usual metaclass layout problems would exist for subclasses because first, it's hard to subclass BTrees anyway and not recommended, and second, BTrees already have a custom layout and the metaclass won't change that.

jamadden:
Just like on the Python clasess. Fixes #166

This takes a metaclass at the C level, but it's a very simple one.

In addition to consistency, and letting the sizes be customized for an entire application, this has some nice properties:

- We can optimize some zope.interface object storage
- We delete the DEFAULT_MAX_*_SIZE macros. Now there's only one source of truth for those, whether in C or Python: _datatypes.py

In addition, I was able to make a small optimization for `__slotnames__`. Previously it was computed (to be empty) and then
discarded over and over (every time you pickled or deactivated an object); now it is properly cached. This won't affect any subclasses.


jamadden:
Two other points:

1) This shouldn't affect existing pickles or older versions reading these pickles.
2) If you do something silly to the `max_*` attribute, you'll get a silly `TypeError`, just as you do in the Python version now.

jamadden:
In testing a large, busy application, increasing the bucket sizes definitely helps scalability. For deployed apps using, e.g., zope.index, it will be much simpler (and faster) to change the bucket sizes globally rather than updating all objects to use [a different BTree family](https://github.com/NextThought/nti.zodb/blob/master/src/nti/zodb/btrees.py) and rebuilding all the data structures.

src/BTrees/BTreeTemplate.c
```diff
@@ -2496,8 +2504,79 @@ static PyNumberMethods BTree_as_number_for_nonzero = {
...
+        if (PyErr_Occurred()) {
+            return -1;
+        }
+        return 0;
+    }
+    PyErr_Format(
+        PyExc_TypeError,
+        /* distinguish the error message from what type would produce */
+        "BTree: can't set attributes of built-in/extension type '%s'",
+        type->tp_name);
```
d-maurer, line 2538: We can set some attributes. I suggest "cannot set attribute ${name} on type ${type}".

d-maurer (approved):

docs/development.rst
```diff
@@ -194,6 +179,30 @@ Macros for Set Operations
...
 
+Datatypes
+=========
+
+There are two tunable values exposed on BTree and TreeSet classes.
+Their default values are found in ``_datatypes.py`` and shared across
+C and Python.
+
+
+``max_leaf_size_str``
```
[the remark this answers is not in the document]
jamadden: Whoops, the `_str` suffix is a mistake. That's what the internal C variable holding the python string "max_leaf_size" is called.

I'm afraid the `max_leaf_size` and `max_internal_size` names are set in stone at this point. That's what they've been called since the feature first existed.

I agree that different names make more sense, which is why I used them in `_datatypes.py` 😄 There are some functions that work with them that also follow those names (e.g., `bucket_size_for_value(value_type)`, so I'd prefer not to change `_datatypes.py`, I think that would be obfuscation. I think it would be better to live with the impedance mismatch at the one place it occurs.

src/BTrees/BTreeTemplate.c
jamadden: Great point! I just [copied what typeobject.c does](https://github.com/python/cpython/blob/e7078ee11667820e3e4585a67d4444d1d469d54e/Objects/typeobject.c#L3401) and stuck "BTree" on the front of it. I'll make a change.

jamadden:
Thanks for the review! I plan to make those textual changes and then merge and release.

PR #169, opened 2021-05-27
Fix setting class attributes on BTree subclasses.

### LINKED ISSUE #168: Regression in 4.9: Subclasses can't use @adapter
Because of the custom metaclass. Previously this code used to work:

```pycon
>>> import BTrees
>>> from zope.component import adapter
>>> from zope.interface import Interface
>>> @adapter(Interface)
... class Subclass(BTrees.family64.OO.BTree):
...     pass
...
  File "<stdin>", line 2, in <module>
  File "//zope/component/_declaration.py", line 33, in __call__
    ob.__component_adapts__ = _adapts_descr(self.interfaces)
TypeError: BTrees: cannot set attribute '__component_adapts__' of immutable type 'Subclass'
```

Apparently there exists code like this in the wild.

jamadden:
This was broken by #166.

Fixes #168. Instead of directly raising the TypeError ourself for an unknown attribute, defer to `PyType_Type.tp_setatro` if we don't know the attribute.

This was a surprise breakage. I'd be grateful if anyone can review this quickly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,10 @@
 4.9.1 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Fix setting unknown class attributes on subclasses of BTrees when
+  using the C extension. This prevented subclasses from being
+  decorated with ``@component.adapter()``. See `issue 168
+  <https://github.com/zopefoundation/BTrees/issues/168>`_.
 
 
 4.9.0 (2021-05-26)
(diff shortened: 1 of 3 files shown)


freddrake (approved):
This looks good to me.

jamadden:
Thank you @freddrake. I am grateful!

The appveyor failure was an unrelated error to resolve `github.com`.

PR #179, opened 2022-06-03
Disable unsafe-math-optimizations in CFLAGS

Nephirus:
Current binary wheels use `-Ofast` optimization option, which (apart from other things) enables flush-to-zero (FTZ) behavior for IEEE-754 subnormals. Unfortunately, FTZ is a global CPU flag, so it affect other running code. Let me add some specific examples:

NumPy emits UserWarning: https://github.com/numpy/numpy/blob/main/numpy/core/getlimits.py#L80

Hypothesis fails on import: https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L211

I have added `-fno-unsafe-math-optimizations` option to manylinux build `CFLAGS`, which disables this kind of standard-violating optimization.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,8 +61,8 @@ env:
   PIP_NO_PYTHON_VERSION_WARNING: 1
   PIP_NO_WARN_SCRIPT_LOCATION: 1
 
-  CFLAGS: -Ofast -pipe
-  CXXFLAGS: -Ofast -pipe
+  CFLAGS: -Ofast -pipe -fno-unsafe-math-optimizations
+  CXXFLAGS: -Ofast -pipe -fno-unsafe-math-optimizations
   # Uploading built wheels for releases.
   # TWINE_PASSWORD is encrypted and stored directly in the
   # github repo settings.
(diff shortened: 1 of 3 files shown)


icemac:
@Nephirus Thank you for your PR. The changes have to be added to https://github.com/zopefoundation/meta/tree/master/config as this is our configuration framework. This way they are not lost when running a configuration update on this package. I'll put it on my list – but it is quite long nowadays.

mgedmin:
Came here because https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html crossed my Twitter feed.  Created https://github.com/zopefoundation/meta/pull/153 as requested.

mgedmin:
I'm having slight trouble updating BTrees from the new metaconfig template, since its tox.ini wants to run tests on Pythons starting with 3.5.  I don't have a Python 3.5 on my current laptop, and my 3.6 is wonky and segfaults.

(Also, I don't remember exactly what commands to run.  I'm trying `cd .../BTrees; .../meta/config/config-package.py .`)

mgedmin:
Okay, https://github.com/zopefoundation/BTrees/pull/184 is what the config-package.py did to the repo.  I had to create a branch and come up with a commit message myself because tox failed on my machine.

icemac:
Superseded by #184.

PR #186, opened 2022-11-08
Work on Python 3.11 failures on MacOS

icemac:
Try the same cure as for Python 3.10.

Let's see if it helps.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -173,8 +173,8 @@ jobs:
       - name: Build BTrees (all other versions)
         if: >
           !startsWith(runner.os, 'Mac')
-          || !startsWith(matrix.python-version, '3.10')
-          || !startsWith(matrix.python-version, '3.11')
+          || !(startsWith(matrix.python-version, '3.10')
+               || startsWith(matrix.python-version, '3.11'))
         run: |
           # Next, build the wheel *in place*. This helps ccache, and also lets us cache the configure
           # output (pip install uses a random temporary directory, making this difficult).
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -154,10 +154,11 @@ jobs:
           pip install -U pip
           pip install -U setuptools wheel twine cffi
 
-      - name: Build BTrees (Python 3.10 on MacOS)
+      - name: Build BTrees (Python 3.10 and 3.11 on MacOS)
         if: >
           startsWith(runner.os, 'Mac')
-          && startsWith(matrix.python-version, '3.10')
+          && (startsWith(matrix.python-version, '3.10')
+              || startsWith(matrix.python-version, '3.11'))
         env:
           _PYTHON_HOST_PLATFORM: macosx-11-x86_64
         run: |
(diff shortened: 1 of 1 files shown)


dataflake (approved):
I have suggested a simplification.

I was going to try the same today after many other attempts. There is something really fishy in how the Python 3.11 on GHA is set up that I cannot reproduce locally.

I'll try this on the remaining packages that need a fix, like zope.security and zope.container and if it works I'll do releases for them (and for BTrees). I think you realize that the solution is just a band-aid. I still think there is a bug in setuptools.

.github/workflows/tests.yml
```diff
@@ -154,10 +154,11 @@ jobs:
...
 
-      - name: Build BTrees (Python 3.10 on MacOS)
+      - name: Build BTrees (Python 3.10 and 3.11 on MacOS)
         if: >
           startsWith(runner.os, 'Mac')
-          && startsWith(matrix.python-version, '3.10')
+          && (startsWith(matrix.python-version, '3.10')
+              || startsWith(matrix.python-version, '3.11'))
         env:
           _PYTHON_HOST_PLATFORM: macosx-11-x86_64
```
dataflake, line 163:
```suggestion
          _PYTHON_HOST_PLATFORM: macosx-10.15-x86_64
```

icemac:
> I think you realize that the solution is just a band-aid.

Totally. I am not sure where the problem is. We set an environment variable suggesting a newer MacOS version. It is still strange to me why this helps. Maybe cibuildwheel already does the necessary settings for us.

.github/workflows/tests.yml
icemac: Why 10.15?

.github/workflows/tests.yml
```diff
@@ -154,10 +154,11 @@ jobs:
           pip install -U pip
           pip install -U setuptools wheel twine cffi
 
-      - name: Build BTrees (Python 3.10 on MacOS)
+      - name: Build BTrees (Python 3.10 and 3.11 on MacOS)
         if: >
           startsWith(runner.os, 'Mac')
-          && startsWith(matrix.python-version, '3.10')
+          && (startsWith(matrix.python-version, '3.10')
+              || startsWith(matrix.python-version, '3.11'))
```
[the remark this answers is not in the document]
icemac: This makes the code easier but seeing `3.1` shakes me.

.github/workflows/tests.yml
dataflake: See test failure in GHA action "test (3.10, macos-latest)". There's actually more to do. The failure in "3.11 (macos-latest)" tells you there's more variables to be set so x86_64 gets built and used.

dataflake:
> > I think you realize that the solution is just a band-aid.
> 
> Totally. I am not sure where the problem is. We set an environment variable suggesting a newer MacOS version. It is still strange to me why this helps. Maybe cibuildwheel already does the necessary settings for us.

I have tried the cibuildwheel experiment on BTrees. It takes more than 6 hours to run, which is totally inacceptable. Besides, I think that experiment goes too far. We do not need eggs for all combinations of architecture and OS under the sun. The effort to build those far exceed any benefit because those users would otherwise just pull the source distribution and quickly build an egg themselves.

Comparing cibuildwheel and the macos-latest VMs used by GHA is hard. The environment is quite different. cibuildwheel uses qemu to set up a VM and then installs the official binary installers (the `.pkg` files for macOS) from python.org.

.github/workflows/tests.yml
dataflake: Looks like the issue was in the condition you already fixed instead

dataflake:
This seems to be fine and I have verified that running meta/config with the branch from PR https://github.com/zopefoundation/meta/pull/168 produces the same GHA configuration file, so I am turning this into a real PR.

dataflake:
Ready to merge, and then I would create a new bugfix release so we get the eggs

icemac:
Thank you for reviewing this PR. 😃

@dataflake I do _not_ plan to create the release myself.

PR #188, opened 2022-12-19
Fix GHA-Tests

icemac:
Try out omitting to install faulthandler.
The repository was archived and the packages was deleted from PyPI 😭 
See https://github.com/vstinner/faulthandler


icemac:
Will be superseded by dropping Python 2 support.

PR #190, opened 2023-01-16
Try to fix installing of 4.11.x on Python 2.7

icemac:
* Remove no longer existing faulthandler.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,7 +52,6 @@ env:
   PYTHONUNBUFFERED: 1
   PYTHONDONTWRITEBYTECODE: 1
   PYTHONDEVMODE: 1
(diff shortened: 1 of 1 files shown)


icemac:
The problem I saw was that BTrees 4.10.0 did not install using `zc.buildout` and Python 2.7 on a linux machine. (There is no wheel for linux and Python 2.7 for newer releases.)
The problem seems to be that `BTrees` has `persistent` in its `setup_requires`. Lately  `persistent 5.0` was released. It removed Python 2 compatibility. Although is correctly marked with `python_requires='>=3.7'`, `zc.buildout` seems to ignore this and it also seems to ignore a version pin set for `persistent`.
I solved the problem by using `pip` as installer. It uses a Python 2 compatible version of `persistent` and everything is fine.

PR #198, opened 2024-02-26
Isort imports + fix Sphinx config.

icemac:


src/BTrees/check.py
```diff
@@ -11,6 +11,7 @@
 # FOR A PARTICULAR PURPOSE
 #
 ##############################################################################
+# isort: skip_file
```
dataflake, line 14: Was this file unfixable? Just wondering.

dataflake (approved):

src/BTrees/check.py
icemac: I did not want isort to touch the groups of imports having a comment each. There is no way to have comments between imports in isort, if I understood correctly.

icemac:
Thank you for reviewing this PR. 😃

PR #205, opened 2024-05-28
Configuring for c-code:  drop Python 3.7 support

tseaver:


tseaver:
@dataflake
> This is missing changes in setup.py to remove the 3.7 trove classifier and adjust the python_requires value.

Hmm, I'm not quite sure how the `meta/config-package.py` bits work for that.  I'll try rerunning against the PR branch, and make them manually if needed.

dataflake (approved):

PR #206, opened 2024-06-02
refactor: scrub Python2 compatibility fossils

tseaver:
- Disentangle checks for needed `long_long`/`ulong_long` helpers. Silences compiler warning for unused functions.
- Replace Py2 compatibility macros w/ Python APIs.
- Remove the local `_compat.h` header where they were defined, replacing includes of it everwhere with `#include "Python.h"`.
  Note that the unrelated `include/persistent/persistent/'compat.h` header in the  directory is still in place.  No C header or module imports it directly, and none uses any macros defined in it indirectly.

This work is in preparation for adding PEP 489 multi-phase module initialization, including compatibility with the changes to `persistent` in https://github.com/zopefoundation/persistent/pull/204. bit does not yet expect (or require) that PR to be merged.


davisagli (approved):

tseaver:
I don't grok the Coveralls failures, but they appear to be complaining that `master` (not just this PRs branch) has dropped statement and / or branch coverage since the merge of PR #198.

davisagli:
@tseaver It's not the first time I've seen confusing reports from coveralls

PR #211, opened 2024-08-26
Apply latest meta/config templates

dataflake:


tseaver (commented):
As with the equivalent `persistent` PR, I don't have a particular issue with the changes, but feel that applying new linting should *not* be done while there are unmerged "substantive" PRs (e.g, #208, #209).

dataflake:
> As with the equivalent `persistent` PR, I don't have a particular issue with the changes, but feel that applying new linting should _not_ be done while there are unmerged "substantive" PRs (e.g, #208, #209).

If your concern is merging issues then I don't see the problem - the linting only affects Python files, which your PRs don't touch at all with one small exception.

src/BTrees/tests/common.py
```diff
@@ -2867,8 +2870,8 @@ def setUp(self):
         super().setUp()
         _skip_if_pure_py_and_py_test(self)
 
-        rawAkeys = [1,    3,    5, 6]
-        rawBkeys = [   2, 3, 4,    6, 7]  # noqa #201
+        rawAkeys = [1, 3, 5, 6]
+        rawBkeys = [2, 3, 4, 6, 7]  # noqa #201
```
icemac, line 2874:
```suggestion
        rawBkeys = [2, 3, 4, 6, 7]
```

icemac (approved):

